PR #5, opened 2017-09-29
added multiple-instace learning example for the Musk1 

pevnak:
Example of multiple-instance learning using neural networks as described in "Using Neural Network Formalism to Solve Multiple-Instance Problems, Tomas Pevny, Petr Somol", https://arxiv.org/abs/1609.07257

I am happy to know, what we did wrong.


MikeInnes:
This looks great to me.

It depends on a couple of patches you've made to Flux itself; I'll wait until those are sorted to merge this (just so it doesn't confuse newer users).

pevnak:
Great,

I was thinking that this will also show some people, how to extend Flux by new layers / model.

Tomas

MikeInnes:
Absolutely!

pevnak:
This is outdated, as this can be now easily solved using Mill.jl extension of Flux.

PR #9, opened 2017-10-26
simple MNIST Autoencoder

iblis17:


MikeInnes:
This is a cool example, thanks! I just want to simplify things a bit and I'll merge.

MikeInnes:
Hope you don't mind that I moved things around a bit here, let me know if I missed anything.

I've removed the model-saving code, because rather than having this on a per-model basis I think we need to solve this in Flux itself and make it easy everywhere. Hopefully will be able to do that before too long.

mnist/autoencoder.jl
```diff
@@ -0,0 +1,46 @@
...
+x ./= 255
+
+# Partition into batches of size 1000
+data = [(x[:,i],) for i in partition(1:60_000, 1000)]
+
+N = 32 # Size of the encoding
+
+m = Chain(
+  Dense(28^2, N, relu), # Encoder
+  Dense(N, 28^2, relu)) # Decoder
```
iblis17: The reason why I separate the encoder and decoder as different variables is that I want user store them (into JLD/serialized file), sometimes the useful part is encoder, sometime is the decoder.

MikeInnes: Makes sense. You can still do this as `m[1], m[2]`, though it might be clearer to use the names -- feel free to make that patch!

PR #10, opened 2017-11-08
fix optimizer in char-rnn

kleinschmidt:
eta is an optional argument, not a kw argument


MikeInnes:
Thanks, I think this changed recently.

PR #14, opened 2018-01-01
Add model for action recognition in videos

americast:
@MikeInnes A model of the Single Frame Multiresolution model as given in the paper [Large-scale Video Classification with Convolutional Neural Networks](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/42455.pdf) implemented with Flux.

Gramercy.


americast:
Local Response Normalisation should be a part of Flux I guess, for proper backprop.

MikeInnes:
This is cool, and a nice demo of how to build this kind of model. Hope you don't mind that I pushed some small style tweaks.

How are gradients working right now? We definitely shouldn't be differentiating through scalars here as it'll be too slow – and besides, it's better to use only constructs that Flux supports. So yeah, we should have local response normalisation in Flux.

The ``run(`git clone ...`)`` step isn't ideal as it won't work out-of-the box on windows. But I can put the data set up somewhere as a zip file easily enough.

PR #17, opened 2018-01-19
Update char-rnn.jl

rajraomichigan:
should be stop = onehot("_", alphabet)
and not stop = onehot('_  ', alphabet)
double quotes as opposed to single quotes

### DIFF
--- a/char-rnn/char-rnn.jl
+++ b/char-rnn/char-rnn.jl
@@ -12,7 +12,7 @@ isfile("input.txt") ||
 text = collect(readstring("input.txt"))
 alphabet = [unique(text)..., '_']
 text = map(ch -> onehot(ch, alphabet), text)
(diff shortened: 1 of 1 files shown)


rajraomichigan:
should be double quotes instead of single quotes

MikeInnes:
Why?

rajraomichigan:
It didn’t work the way it is listed :)
________________________________

MikeInnes:
Can you clarify "doesn't work"? Are you getting some error?

It's working for me, and notice that the alphabet is defined as `alphabet = [unique(text)..., '_']`, so `" "` will not be found in the alphabet correctly.

rajraomichigan:
I was getting an error when I tried it (while Alan was in SG)..must have been a copy/paste bug ...all set! Thanks!


________________________________

PR #18, opened 2018-02-01
mnist/autoencoder: add more comment about single encoder/decoder

iblis17:
address: https://github.com/FluxML/model-zoo/pull/9#discussion_r160960134


MikeInnes:
:+1: Thanks!

PR #21, opened 2018-02-08
VAE Implementation

willtebbutt:
(Approximately) reproduces the (binarized) MNIST example in [1]. Uses the same architecture as that proposed in the paper, but uses Adam rather than Adagrad, Xavier init. rather than sampling each parameter's initial value from a Gaussian with zero mean and variance of 0.01, and a broader prior on the model parameters.

I suspect my iterator implementation is unnecessary, so I would appreciate advice here if this is the case.

[1] - Kingma, Diederik P., and Max Welling. "Auto-encoding variational bayes." arXiv preprint arXiv:1312.6114 (2013).

### DIFF
--- a/mnist/vae.jl
+++ b/mnist/vae.jl
@@ -61,9 +61,9 @@ sample(M::Int=1) = rand.(Bernoulli.(f(z.(zeros(Dz, M), zeros(Dz, M)))))
 
 ################################# Learn Parameters ##############################
 
-evalcb = throttle(() -> @show(-L̄(X[:, rand(1:60000, M)])), 5);
+evalcb = throttle(() -> @show(-L̄(X[:, rand(1:60000, M)])), 30);
 opt = ADAM(vcat(params(A), params(μ), params(logσ), params(f)));
-@progress for i = 1:10
+@progress for i = 1:15
   info("Epoch $i")
-  Flux.train!(X->-L̄(X) + 0.5 * sum(x->sum(x.^2), params(f)), data, opt, cb=evalcb)
+  Flux.train!(X->-L̄(X) + 1e-2 * sum(x->sum(x.^2), params(f)), data, opt, cb=evalcb)
 end
--- a/mnist/vae.jl
+++ b/mnist/vae.jl
@@ -1,20 +1,8 @@
-using Flux, Flux.Data.MNIST, PyPlot
+using Flux, Flux.Data.MNIST, MLDataUtils
 using Flux: throttle, params
 using Juno: @progress
 
 rng = MersenneTwister(1234567)
-
-# Add an iterator to simply run over the data set.
-struct MinibatchIterator{TX<:AbstractMatrix}
-  X::TX
-  M::Int
-end
-Base.start(d::MinibatchIterator) = 1
-function Base.next(d::MinibatchIterator, p::Int)
-  p′ = min(p + d.M, size(d.X, 2))
-  return (view(d.X, :, p:p′-1),), p′
-end
-Base.done(d::MinibatchIterator, p::Int) = p == size(d.X, 2)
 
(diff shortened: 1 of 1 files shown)


mnist/vae.jl
```diff
@@ -0,0 +1,69 @@
...
+
+################################# Define Model #################################
+
+# Latent dimensionality, # hidden units.
+Dz, Dh = 5, 500
+
+# Components of recognition model / "encoder" MLP.
+A, μ, logσ = Dense(28^2, Dh, tanh), Dense(Dh, Dz), Dense(Dh, Dz)
+g(X) = (h = A(X); (μ(h), logσ(h)))
+z(μ, logσ) = μ + exp(logσ) * randn(rng)
```
CarloLucibello: this I think should be
```julia
z(μ, logσ) = μ .+ exp.(logσ) .* randn(rng, Dz)
```

mnist/vae.jl
```diff
@@ -0,0 +1,69 @@
...
+####################### Define ways of doing things with the model. #######################
+
+# KL-divergence between approximation posterior and N(0, 1) prior.
+kl_q_p(μ, logσ) = 0.5 * sum(exp.(2 .* logσ) + μ.^2 - 1 .+ logσ.^2)
+
+# logp(x|z) - conditional probability of data given latents.
+logp_x_z(x, z) = sum(logpdf.(Bernoulli.(f(z)), x))
+
+# Monte Carlo estimator of mean ELBO using M samples.
+L̄(X) = ((μ̂, logσ̂) = g(X); (logp_x_z(X, z.(μ̂, logσ̂)) - kl_q_p(μ̂, logσ̂)) / M)
```
CarloLucibello: I don't know how much is important in this case, but generating random numbers one at a time with `z.(μ̂, logσ̂)` could be noticeably more slow than calling `randn(rng, Dz)`

CarloLucibello:
nice! very beautiful and compact implementation

CarloLucibello:
maybe MLDataPattern.jl's `eachbatch` method could be used instead of the custom iterator

mnist/vae.jl
willtebbutt: Interesting. Why would you expect this to be the case? I just ran some performance tests locally and didn't find it to be an issue.

CarloLucibello: you are right, I remember reading something about that, but I tested it as well and doesn't change much

willtebbutt:
Thanks for the pointer towards `eachbatch`. I went with `batchview` instead as it avoids copying of data which would have been unnecessary in my situation.

MikeInnes:
I agree with Carlo, this is really nice!

I added a sample viewer, and hope you don't mind that I simplified things a bit; I've used a simpler data iterator just to avoid the outside dependency.

For pedagogical purposes it'd be nice to cut away anything that isn't absolutely necessary here; do we need the custom RNG? Would also be great to get the pdf function into Distributions.jl to throwing people off.

willtebbutt:
The simplifications look good. I have mixed feelings about the choice of iterator: on the one hand I agree that it's good to avoid the outside dependency, but on the other I'm not entirely sure that it's good practice to be making lots of copies of the data unnecessarily. Happy to disregard this for pedagogy though.

Happy to dispense with the custom RNG seeding - I've just got used to doing it but it shouldn't really matter here. I agree regarding the logpdf, but I would be surprised if they would want it given that it takes roughly twice as long to compute as their current implementation. Doing this would also save Flux from requiring it's own custom binary cross-entropy implementation, which would be handy.

MikeInnes:
> I would be surprised if they would want it given that it takes roughly twice as long to compute as their current implementation.

It might be possible to switch implementations around the unstable area, and still be mostly fast while also being more correct. I think the community generally errs on that side where possible.

Anyway, let's go with this for now and we hopefully won't need the overload in future. Thanks a lot @willtebbutt!

PR #22, opened 2018-02-11
Parity of bit strings

domluna:
Hey, did this for fun in an effort learn Flux, super cool how straightforward it is to use RNNs with variable length strings.

Basically took the lang-detection model and tweaked it a bit so I understand if you don't wanna merge this.


MikeInnes:
I like it! Thanks for the nice readme and explanation as well.

PR #26, opened 2018-02-20
[DO NOT MERGE] try LearningStrategies out

iblis17:


iblis17:
I add two commits, leverage `MaxIter` and `Tracer` from the package.
I think make LearningStrategies more versatile is a good direction, e.g. port function `throttle` as a `LearningStrategy`; add more and more handy callback utility, maybe someone can benifit from it.

Issue #30, opened 2018-03-17
GSOC Proposal : Computer Vision Networks


avik-pal:
I wish to implement the following network architectures for my GSOC. They are listed in the order of implementation
- [x] VGG16
- [x] VGG19
- [x] ResNet50
- [x] Xception
- [x] Inception V3
- [x] Inception ResNet V2
- [x] MobileNet
- [x] DenseNet
- [ ] NASNet
- [ ] R-FCN
- [ ] Fast R-CNN
- [ ] Faster R-CNN
- [ ] RetinaNet
- [ ] Mask R-CNN (*)

@MikeInnes how does this sound? I wish to test these implementation on the tiny-imagenet-200 dataset or the CIFAR-10

MikeInnes:
GSoC is in progress now, so closing these issues.

EMCP:
anyway I can help with the Faster R-CNN ?  Is someone already working on this?

I am wanting to use transfer learning with Flux, and ween off PyTorch.. but it's tough without a step by step way.. so thinking maybe I can help someone else who may already be in progress?

MikeInnes:
@avik-pal may be able to help or point you to someone working on this, but you might also be interested in signing up to the Julia slack for this kind of thing.

avik-pal:
@dhairyagandhi96 has an implementation of MaskRCNN which is essentially just an extension of Faster RCNN.

PR #33, opened 2018-03-18
Added VGG16 and VGG19 model for CIFAR10 Dataset

avik-pal:


avik-pal:
#30

cifar10/vgg19.jl
```diff
@@ -0,0 +1,85 @@
...
+conv5_2 = Conv((3,3), 512=>512, relu, pad=(1,1), stride=(1,1))
+conv5_3 = Conv((3,3), 512=>512, relu, pad=(1,1), stride=(1,1))
+conv5_4 = Conv((3,3), 512=>512, relu, pad=(1,1), stride=(1,1))
+
+# Building the model
+
+m = Chain(
+	conv1_1,
+    conv1_2,
+    x -> maxpool(x, (2,2)),
```
CarloLucibello: what about 
```julia
Chain( 
    Conv((3,3), 3=>64, relu, pad=(1,1), stride=(1,1)),
    Conv((3,3), 64=>64, relu, pad=(1,1), stride=(1,1)),
.....
```
to save some space?

avik-pal: Made the changes that have been suggested

MikeInnes:
This is great. I've just tried to simplify it a bit; let me know if it's important to keep the batchnorm models separate, but it seems reasonable to just have a sensible default here.

Looks like this depends on a Metalhead release, and the CIFAR10 download is not working there for me, so we'll get that figured out first.

avik-pal:
Keeping one model is fine. At that time I was not being able to train the `BatchNorm` ones due to speed constraints but now it should be fine.
As for the `CIFAR10` I am looking into that

MikeInnes:
Note the small change to `getarray`; you want `Images.channelview` to avoid writing that by hand.

Otherwise this is great, thanks a lot @avik-pal!

PR #34, opened 2018-03-24
Switch autoencoder example to use leakyrelu

amellnik:
Using normal relu causes a small portion of the pixels in the reconstructed image to always be black.  It's a bit subtle with normal MNIST but becomes clear with other datasets like F-MNIST.  You can see this effect in the upper-left of the last four images below:

image


MikeInnes:
makes sense, thanks!

PR #35, opened 2018-03-25
Residual Network

avik-pal:
Added a sample Residual Network which is going to be trained on CIFAR10. The code can be used as a building block to build deeper Residual Networks. 
#30


MikeInnes:
Note that the way you're currently doing this is inefficient since you'll end up calling many of the sub-models twice with the same input.

[Here's](https://gist.github.com/MikeInnes/3aabf8f0db3e3477d50e31572670c385) a version I put together which supports building multiple resnet variants.

MikeInnes:
This is looking great, but one improvement might be to have a test harness that uses Metalhead's ImageNet dataset.

vision/imagenet/resnet.jl
```diff
@@ -0,0 +1,127 @@
...
+    ResidualBlock(filters, [(i,i) for i in kernels], [(i,i) for i in pads], [(i,i) for i in strides], shortcut)
+end
+
+function (block::ResidualBlock)(input)
+    local value = copy.(input)
+    for i in 1:length(block.conv_layers)-1
+        value = relu.((block.norm_layers[i])((block.conv_layers[i])(value)))
+    end
+    relu.(((block.norm_layers[end])((block.conv_layers[end])(value))) + block.shortcut(input))
+end
```
staticfloat: Let's avoid unnecessary `copy()`'s, especially since residual networks are already quite high on memory requirements.  Maybe something like this:

```
function (block::ResidualBlock)(input)
    value = block.norm_layers[1](block.conv_layers[1](input))
    for i in 2:length(block.conv_layers)
        value = block.norm_layers[i])(block.conv_layers[i](relu.(value)))
    end
    return relu.(value .+ block.shortcut(input))
end
```

staticfloat: Let's avoid unnecessary `copy()`'s, especially since residual networks are already quite high on memory requirements.  Maybe something like this:

```
function (block::ResidualBlock)(input)
    value = block.norm_layers[1](block.conv_layers[1](input))
    for i in 2:length(block.conv_layers)
        value = block.norm_layers[i])(block.conv_layers[i](relu.(value)))
    end
    return relu.(value .+ block.shortcut(input))
end
```

vision/imagenet/resnet.jl
```diff
@@ -0,0 +1,127 @@
...
+    shortcut = Chain(Conv((3,3), filters÷2=>filters, pad = (1,1), stride = (2,2)), BatchNorm(filters))
+    ResidualBlock([filters÷2, filters, filters], [3,3], [1,1], [1,2], shortcut)
+end
+
+# Function to generate the residual blocks used for ResNet50, ResNet101 and ResNet152
+function Bottleneck(filters::Int, downsample::Bool = false, res_top::Bool = false)
+    if(!downsample && !res_top)
+        return ResidualBlock([4 * filters, filters, filters, 4 * filters], [1,3,1], [0,1,0], [1,1,1])
+    elseif(downsample && res_top)
+        return ResidualBlock([filters, filters, filters, 4 * filters], [1,3,1], [0,1,0], [1,1,1], Chain(Conv((1,1), filters=>4 * filters, pad = (0,0), stride = (1,1)), BatchNorm(4 * filters)))
```
staticfloat: This line is pretty long, can you prettify this up at all?

avik-pal:
This was added before Metalhead. Since we have shifted it there and the only difference is the ability to train on Imagenet, can't we simply demonstrate using Metalhead's resnet and training using that?

staticfloat:
Yeah, why don't you just slice out the code here that is also in `Metalhead` and instead just instantiate the Metalhead model and train it here.  That would be good to do.

PR #36, opened 2018-04-09
Prioritized  Experience Replay DQN

tejank10:
Implemented DQN with PER.
Ref: 
- https://arxiv.org/pdf/1511.05952.pdf
- https://github.com/MorvanZhou/Reinforcement-learning-with-tensorflow/blob/master/contents/5.2_Prioritized_Replay_DQN/RL_brain.py

### DIFF
--- a/prioritized-dqn/memory.jl
+++ b/prioritized-dqn/memory.jl
@@ -1,0 +1,61 @@
+include("sum-tree.jl")
+
+using Distributions
+
+mutable struct Memory  # stored as ( s, a, r, s_ ) in SumTree
+    ϵ::Float32  # small amount to avoid zero priority
+    α::Float32  # [0~1] convert the importance of TD error to priority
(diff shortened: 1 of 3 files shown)


MikeInnes:
This looks awesome. Unfortunately on my setup it throws an error:
```julia
MethodError: Cannot `convert` an object of type Flux.Tracker.TrackedReal{Float64} to an object of type Float64
```
Let me know if you can't reproduce this and we'll narrow it down.

MikeInnes:
Once you get it going this is a pretty cool demo, because you can see the thing learning in real time.

Is it possible to get it working on the GPU?

Is there any way to skip ahead a few iterations without the visualisation? Just to see how the model performs after the full training.

tejank10:
Yes, I'll get it on GPU.
Skipping ahead is possible by not rendering the frams, but I am not sure if that'll improve the training speed. I'll try that.

PR #37, opened 2018-04-17
char-rnn on gpu

gustafsson:
Takes 5 minutes on a GPU, vs, 60 minutes on four CPUs. Both train the loss to about 80-85 and produce similar sample output.

### DIFF
--- a/char-rnn/char-rnn.jl
+++ b/char-rnn/char-rnn.jl
@@ -1,5 +1,6 @@
+using CuArrays
 using Flux
-using Flux: onehot, argmax, chunk, batchseq, throttle, crossentropy
+using Flux: onehot, argmax, chunk, batchseq, throttle, logitcrossentropy
 using StatsBase: wsample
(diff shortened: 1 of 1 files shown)


gustafsson:
`crossentropy` uses `log` which isn't supported by CuArrays (https://github.com/JuliaGPU/CuArrays.jl/issues/71) so I changed it to use `logitcrossentropy` instead.

gustafsson:
Changed back from `logitcrossentropy` to `crossentropy` since https://github.com/FluxML/Flux.jl/commit/8f73dc6e148eedd11463571a0a8215fd87e7e05b. Got the same performance results and similar output.

DilumAluthge:
Any update on merging this PR? @gustafsson @MikeInnes

gustafsson:
awaiting review, don't know if there's much else I can do

MikeInnes:
Sorry for the wait, but thanks for the great patch!

PR #39, opened 2018-05-06
MADE

tejank10:
MADE is Masked AutoEncoder for Density Estimation. @MikeInnes @jekbradbury


MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+    return # only a single seed, skip for efficiency
+  end
+
+  L = length(made.hidden_sizes)
+
+  # fetch the next seed and construct a random stream
+  rng = MersenneTwister(made.seed)
+  made.seed = (made.seed + 1) % made.num_masks
+
+  # sample the order of the inpdimensionsuts and the connectivity of all neurons
```
jekbradbury: typo

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+function (made::MADE)(x)
+  made.net(x)
+end
+
+# ------------------------------------------------------------------------------
+
+#Getting data. The data used here is binarized MNIST dataset
+
+X = npzread("/path/to/your/data.npy")
+X = X'
```
jekbradbury: Punning on adjoint is not really recommended, right? And deprecated in 0.7, so maybe use permutedims?

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+  seed::UInt  # for cycling through num_masks orderings
+
+  m::Dict
+
+  function MADE(in::Integer, hs, out::Integer, nat_ord::Bool, num_masks = 1)
+    # define a simple MLP neural net
+    hs = push!([in], hs...)
+    layers = [MaskedDense(hs[i], hs[i + 1]; σ = relu) for i = 1:length(hs) - 1]
+
+    net = Chain(layers..., MaskedDense(hs[end], out))
```
jekbradbury: Couldn't this have a special case for the first and last layers (rather than just the last one), allowing you to avoid the `push!` in line 57?

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+
+Flux.treelike(MaskedDense)
+
+function (a::MaskedDense)(x)
+  a.σ.(a.mask .* a.W * x .+ a.b)
+end
+
+function set_mask(a::MaskedDense, mask)
+  a.mask = mask
+end
```
jekbradbury: Unless there's validation that you need to do this probably doesn't need to be a setter method and could just be replaced at it so call locations by directly setting the attribute?

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+struct MaskedDense{F,S,T,M}
+  # same as Linear except has a configurable mask on the weights
+  W::S
+  b::T
+  mask::M
+  σ::F
+end
+
+function MaskedDense(in::Integer, out::Integer; σ = identity)
+  return MaskedDense(param(glorot_uniform(out, in)), param(zeros(out)), ones(out, in), σ)
```
jekbradbury: This could probably be modified to be more type-generic? Would be nice to be able to support GPU and/or Float32

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+  a.mask = mask
+end
+
+# ------------------------------------------------------------------------------
+
+mutable struct MADE
+  nin::Integer
+  nout::Integer
+  hidden_sizes::Array{Integer, 1}
+  net::Chain
```
jekbradbury: These should probably be concretely typed (`Int` and either `Vector{Int}` or a tuple)

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+
+function update_masks(made::MADE)
+  if made.m != Dict() && made.num_masks == 1
+    return # only a single seed, skip for efficiency
+  end
+
+  L = length(made.hidden_sizes)
+
+  # fetch the next seed and construct a random stream
+  rng = MersenneTwister(made.seed)
```
jekbradbury: Is this reinitialization going to be slow?

MADE/made.jl
```diff
@@ -0,0 +1,140 @@
...
+  nout::Integer
+  hidden_sizes::Array{Integer, 1}
+  net::Chain
+
+  # seeds for orders/connectivities of the model ensemble
+  natural_ordering::Bool
+  num_masks
+  seed::UInt  # for cycling through num_masks orderings
+
+  m::Dict
```
jekbradbury: Maybe give this dict a concrete type for clarity about what it's going to hold?

MADE/made.jl
tejank10: Wouldn't it be an issue if `hs = []`? In that case, `Chain(MaskedDense(in, hs[1]), layers..., MaskedDense(hs[end], out))` gives `BoundsError`

MADE/made.jl
tejank10: Could you please explain a bit more on that? Because `in` and `out` have to be integers, `MaskedDense` has been modeled after `Dense`.

MikeInnes:
Can you update this to use `Flux.MNIST` rather than the numpy file? We'll need this to work out of the box.

Also [this line] seemed sketchy, given that maskeddense is immutable, what's going on there?

PR #41, opened 2018-05-08
[WIP] Depthwise Seperable Convolutions for Xception and MobileNet

avik-pal:
Adds support for Depthwise Convolution Layer. The layer in its current form can be swapped in for any Convolution Layer.

### DIFF
--- a/imagenet/seperableconv.jl
+++ b/imagenet/seperableconv.jl
@@ -1,0 +1,44 @@
+using Flux
+using Flux: initn
+using NNlib: conv
+
+struct DepthwiseConv
+    σ
+    weight
+    bias
+    stride
+    pad
+end
+
+Flux.treelike(DepthwiseConv)
+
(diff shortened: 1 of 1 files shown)


datnamer:
Shouldn't this go in Flux.jl or NNlib?

avik-pal:
This is actually for the Xception Network so once that is done and the layer is properly tested I will send in a PR in Flux.jl.

jekbradbury:
Nit: it's spelled "separable"

avik-pal:
NOTE: The residualblock.jl code is same as the one in #35

avik-pal:
Xception Net is ready however the Depthwise Separable Convolution Operations are too slow for large number of channels

avik-pal:
Moving the implementation of Depthwise Convolutions to FluxML/NNlib.jl#42

MikeInnes:
What's the status of this now?

avik-pal:
Waiting for the layer to be merged in Flux. Will update the API after that. This version works but is slower than that one(in Flux).

PR #43, opened 2018-05-15
Dueling Double DQN on Pong

tejank10:
Added Dueling DDQN model


MikeInnes:
Closing this in favour of the flux-baselines model, for now.

PR #45, opened 2018-06-02
Added Inception Nets for Places 2 Dataset

avik-pal:


dhairyagandhi96:
Bump!

avik-pal:
I think we can remove this model from here, and have the new Metalhead API for training this?

dhairyagandhi96:
I think that would be fine
cc @MikeInnes

PR #50, opened 2018-08-14
Add framewise speech recognition network

maetshju:
This is a submission of one of the neural networks I worked on, and the one that actually showed some reasonable level of performance if trained long enough. It does not use the fancy CTC loss function, but it outperforms any of the CTC ones that I've attempted.

### DIFF
--- a/audio/speech-blstm/00-data.jl
+++ b/audio/speech-blstm/00-data.jl
@@ -24,28 +24,28 @@ const PHONE_TRANSLATIONS = translations
 
 # Make dictionary to perform class folding
 const FOLDINGS = Dict(
-    "ao" => "aa",
-    "ax" => "ah",
-    "ax-h" => "ah",
-    "axr" => "er",
-    "hv" => "hh",
-    "ix" => "ih",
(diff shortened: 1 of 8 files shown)


MikeInnes:
Cool -- could you just add a project/manifest file / make sure this works on 0.7?

A README with a bit of background / more instructions would also be great.

maetshju:
I think that all sounds great. I have a conference coming up this week that is keeping me busy, but I should be able to get to this after that.

maetshju:
I've updated and tested the code to work with 0.7/1.0, added a readme, and added the toml files. I was a little unsure of what to do with the toml files, since the MFCC package still needs either https://github.com/JuliaDSP/MFCC.jl/pull/13 or https://github.com/JuliaDSP/MFCC.jl/pull/12 to be merged, but it looks like Pkg3 wrote the toml files to use the branches from the PRs.

MikeInnes:
Perfect, thanks! It would be nice if we could eventually use DataDeps or similar for the TIMIT data, but this is fine for now.

PR #52, opened 2018-08-22
Update phonemes example

ishai1:


ishai1:
[Fixed here](https://github.com/FluxML/model-zoo/pull/57)

PR #54, opened 2018-09-08
 Fizz Buzz in Flux

Drvi:
Add a toy example inspired by "Fizz Buzz in Tensorflow" blog by Joel Grus


MikeInnes:
Nice :) Would you be able to add a project/manifest file to pin packages? Then this is good to go in.

Drvi:
Done! :)

PR #55, opened 2018-09-12
fix housing.jl for Julia 1.0

mcreel:
Simple fixes for Julia 1.0


dhairyagandhi96:
Thanks for this. Some of the changes seem to have been merged already, and the PR comes from an unknown repo, so crediting becomes difficult. Would be happy to take suggestions on this and other models, though. Closing for now.

PR #66, opened 2018-10-18
[WIP] Code for training UNet on Cavana Image Masking Kaggle Competition

avik-pal:
The code is complete and mostly tested. I will remove the WIP tag once I finish training the model. Also will add the Project.toml files and README once training is done.
Also the convtranspose code here is borrowed from the `FastStyleTransfer` and should removed once the ConvTranspose PR is merged.


MikeInnes:
Cool! Hopefully you can use Flux's new ConvTranspose layer before too long.

Also, can you open an issue on why you're not using the Flux training loop? I guess just so you can accumulate the loss, but especially for the model zoo it'd be good to make sure that Flux's built-ins are flexible enough to do this stuff.

avik-pal:
Yeah, I am doing that so that I can accumulate the loss. Is there a way to do that in other way?
For the metric I can use the callbacks.
I will make those changes

avik-pal:
I managed to train the network. It is giving good results even after a few epochs. I will make the updates for using the Flux default `train!`. Also, I figured a way to accumulate the loss and metrics without rewriting the training loop.

I will remove the WIP tag after the `ConvTranspose` PR is merged and I have made the changes to the model here.

amellnik:
What's the status on this PR?  If no one is currently working on it I can update it to the latest Flux.

avik-pal:
@amellnik it would be great if you could do that.

PR #71, opened 2018-11-06
Upgraded lang-detection to Julia 1.0

NTimmons:
model.jl 
- Renamed the normalize_string function to normalize from the Unicode library
- Added 'using Random' for the shuffle function
- Added 'using Statistics' for the mean function

scrape.jl
- Changed matchall to eachmatch from the Cascadia library. Which appears to behave the same and is consistent with the usage guide on the github page for that library (https://github.com/Algocircle/Cascadia.jl)
- moved from sel"some regex" to Selector("some regex") in the match call so that a new using can easily find the Selector method. (I had trouble finding the documentation for the sel prefix)
- IOBuffer has changed. Switched innerText to work on a string instead. Might be less efficient but task is trivial.

(I am fairly new to Julia, let me know if anything is not general practice or a bit wonky!)


dhairyagandhi96:
Thanks for this!

NTimmons:
No problem :)

text/lang-detection/model.jl
```diff
@@ -1,13 +1,17 @@
-using Flux
+using Flux       # Main ML library
 using Flux: onehot, onehotbatch, crossentropy, reset!, throttle
 
+using Unicode    # For 'normalize'
+using Random     # for 'shuffle'
+using Statistics # for 'mean'
```
dhairyagandhi96: Seeing as we are only using one function from these modules, do you think something like this would be better?

```suggestion
using Statistics: mean
```

text/lang-detection/scrape.jl
```diff
@@ -7,17 +10,18 @@ pages = Dict(
...
-rawpage(url) = parsehtml(String(HTTP.get(url).body)).root
-
 function innerText(dom)
-  text = IOBuffer()
+  text = ""
   for elem in PreOrderDFS(dom)
-    elem isa HTMLText && print(text, elem.text)
+        if elem isa HTMLText
+            text = string(text, elem.text)
+        end
```
dhairyagandhi96: We could perhaps retain `IOBuffer` with something like `write(text, elem.text)`?

dhairyagandhi96:
We should also definitely add a `Project.toml` and `Manifest.toml` to this, to be consistent with the rest of the models.

Since we have to convert these to notebooks, we should probably also make the model GPU compatible.

NTimmons:
Moved to: #85 
To update the PR with the correct branch with requested changes.

dhairyagandhi96:
Closing in favour of #122 
We need to get it on the GPUs as well, which will be appreciated

PR #72, opened 2018-11-21
adding epochs to conv mnist classifier

mcognetta:
As discussed on slack, the conv mnist classifier only trained for one epoch. This PR corrects that.

The mlp classifier also only trains for one epoch, but simulates 200 epochs

```
dataset = repeated((X, Y), 200)
evalcb = () -> @show(loss(X, Y))
opt = ADAM(params(m))

Flux.train!(loss, dataset, opt, cb = throttle(evalcb, 10))
```
Should this also be fixed to promote best practices?

This is my first PR here, so any advice is greatly appreciated.

### DIFF
--- a/vision/mnist/conv.jl
+++ b/vision/mnist/conv.jl
@@ -1,5 +1,5 @@
 using Flux, Flux.Data.MNIST, Statistics
-using Flux: onehotbatch, onecold, crossentropy, throttle
+using Flux: onehotbatch, onecold, crossentropy, throttle, @epochs
 using Base.Iterators: repeated, partition
 # using CuArrays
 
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
I think just adding a comment pointing the user to the right way of adding epochs is a good idea, since it is likely to be used interactively. What do you think? 

Thanks! And apologies for the delay

MikeInnes:
That or just include the `@epochs 10 ...` line just after the fast training loop. Good to have it there but yeah, the expectation is that this'll be run in notebooks and such so people will want to do an epoch at a time and inspect the results.

mcognetta:
In either case, I think it should just be consistent between the conv and mlp models. If using `@epochs` is the idiomatic way, then I think that's what should be used. People can then modify that as they wish.

MikeInnes:
I think we have this incorporated into #104. Sorry for the delay in responding but thanks a lot for pushing this forward. Perhaps we can add a commit there to make sure you're attributed.

Issue #73, opened 2018-11-22
kl-divergence formula in vae seems wrong


chengchingwen:
The kl-divergence for normal distribution should be 
[...]

PR #76, opened 2018-12-10
Changed the KL-Divergence formula in vae.jl #73

AbinavRavi:
Changed the KL-divergence formula from log(sigma).^2 to 2*log(sigma)


MikeInnes:
Thanks!

PR #83, opened 2019-01-17
Updated Flux API

dhairyagandhi96:


other/fizzbuzz/fizzbuzz.jl
```diff
@@ -49,6 +50,6 @@ end
 
 # Training
 for e in 0:1000
-    train!(loss, [(X, y)], opt)
+    train!(loss, ps, [(X, y)], opt)
```
MikeInnes: In some of these simple cases I think we should just inline `ps`.

```suggestion
    train!(loss, params(m), [(X, y)], opt)
```

text/treebank/recursive.jl
```diff
@@ -29,8 +29,10 @@ end
...
-opt = ADAM(params(embedding, W, sentiment))
+opt = ADAM()
+ps = params(embedding, W, sentiment)
 evalcb = () -> @show loss(train[1])
 
-Flux.train!(loss, zip(train), opt,
-            cb = throttle(evalcb, 10))
+Flux.train!(loss, ps, zip(train), opt,
+           cb = throttle(evalcb, 10))
+end
```
MikeInnes: What's this `end` for?

vision/mnist/vae.jl
```diff
@@ -47,19 +47,20 @@ modelsample() = rand.(Bernoulli.(f(z.(zeros(Dz), zeros(Dz)))))
...
   @info "Epoch $i"
-  Flux.train!(loss, zip(data), opt, cb=evalcb)
+  Flux.train!(loss, ps, zip(data), opt, cb=evalcb)
 end
 
 
 ################################# Sample Output ##############################
 
-using Images
+#using Images
```
MikeInnes: Why is this getting commented out?

text/treebank/recursive.jl
dhairyagandhi96: Thanks

other/fizzbuzz/fizzbuzz.jl
dhairyagandhi96: Got it

vision/mnist/vae.jl
dhairyagandhi96: This was an artefact of debugging the notebooks generation. Thanks for catching it

KristofferC:
The CUDA versions some of the models (e.g. MNIST conv) doesn't work anymore due to the project files not being updated.

MikeInnes:
Should've caught that. @dhairyagandhi96 can you fix that ASAP?

KristofferC:
Actually, even after updating the packages it doesn't work

```
ERROR: LoadError: GPU compilation of #23(CuArrays.CuKernelState, CUDAnative.CuDeviceArray{Bool,1,CUDAnative.AS.Global}, Base.Broadcast.Broadcasted{Nothing,Tuple{Base.OneTo{Int64}},typeof(==),Tuple{Base.Broadcast.Extruded{CUDAnative.CuDeviceArray{Int64,1,CUDAnative.AS.Global},Tuple{Bool},Tuple{Int64}},Base.Broadcast.Extruded{Array{Int64,1},Tuple{Bool},Tuple{Int64}}}}) failed
KernelError: passing and using non-bitstype argument

Argument 4 to your kernel function is of type Base.Broadcast.Broadcasted{Nothing,Tuple{Base.OneTo{Int64}},typeof(==),Tuple{Base.Broadcast.Extruded{CUDAnative.CuDeviceArray{Int64,1,CUDAnative.AS.Global},Tuple{Bool},Tuple{Int64}},Base.Broadcast.Extruded{Array{Int64,1},Tuple{Bool},Tuple{Int64}}}}.
That type is not isbits, and such arguments are only allowed when they are unused by the kernel.
```

PR #84, opened 2019-01-18
Implemented DQN on CartPole-v0

raghav9-97:
Implemented Deep Q Network on CartPoleV0 environment from Reinforce.jl.

### DIFF
--- a/games/cartpole/CartPole-v0.jl
+++ b/games/cartpole/CartPole-v0.jl
@@ -112,11 +112,6 @@ for e=1:EPISODES
     if length(memory) >= BATCH_SIZE
         exp_replay()
     end
-<<<<<<< HEAD
 end
 
 test()
-
-=======
-end
->>>>>>> parent of 4825afd... Implemented Cartpole gpu version


games/CartPole-v0.jl
```diff
@@ -0,0 +1,94 @@
...
+ϵ = 1.0                 # exploration rate
+ϵ_min = 0.01            # exploration minimum
+ϵ_decay = 0.995         # exploration decay
+
+memory = []             # used to remember past results
+
+#-------------Model Architecture--------------#
+model = Chain(Dense(STATE_SIZE, 24), Dense(24, 24), Dense(24, ACTION_SIZE))
+loss(x, y) = Flux.mse(model(x), y)
+opt = ADAM(params(model), η)
```
tejank10: The optimizer can be updated as per the current API, which does not take params as input.

games/CartPole-v0.jl
```diff
@@ -0,0 +1,94 @@
...
+BATCH_SIZE = 32
+
+γ = 0.95                # discount rate
+η = 0.001               # learning rate
+
+ϵ = 1.0                 # exploration rate
+ϵ_min = 0.01            # exploration minimum
+ϵ_decay = 0.995         # exploration decay
+
+memory = []             # used to remember past results
```
tejank10: Since memory functions like a queue of fixed size, [CircularBuffer](http://juliacollections.github.io/DataStructures.jl/latest/circ_buffer.html#CircularBuffer-1) from DataStructures.jl can be used.

games/CartPole-v0.jl
raghav9-97: I don't think that would work as the definition of ADAM when checked in REPL turns out to be ```ADAM(params, η = 0.001; β1 = 0.9, β2 = 0.999, ϵ = 1e-08, decay = 0)``` where it is necessary to pass params.

games/CartPole-v0.jl
raghav9-97: I will get to it.

games/CartPole-v0.jl
tejank10: The new API is [here](https://github.com/FluxML/Flux.jl/blob/master/src/optimise/optimisers.jl).

MikeInnes:
Looks good, please put it in a `cartpole` folder and add a project/manifest.

dhairyagandhi96:
This looks nice! Would it be possible to get a GPU friendly version up as well?

raghav9-97:
I will surely start working on this asap.

dhairyagandhi96:
Looks good 👍

I think adding a function to see how the trained model performs would be nice.

dhairyagandhi96:
Hmm.. the current model seems to be having a hard time playing the game. Can we perhaps focus a little on getting the model stable. As it stands, we have a hard time cracking 3 digits.

dhairyagandhi96:
Bump

PR #80, opened 2019-01-01
Fixes to CIFAR10 Model

joshua-whittemore:
Force getarray to return CIFAR images as Float64, using float method returns Float32.

cat takes dims as a keyword arg.

PR #89, opened 2019-01-24
Fix issues with the CIFAR10 VGGs

avik-pal:
1. This fix will work only after FluxML/Flux.jl#570 is merged.
2. Metalhead has not been tagged for quite sometime. Automatic data download depends on fixes on the latest master, so I am bumping the version requirement to master for that.


dhairyagandhi96:
Ref: #80

MikeInnes:
Flux is tagged, can you update the manifest to latest versions and test?

avik-pal:
@MikeInnes Made the updates and tested.

PR #91, opened 2019-01-25
Add neural SDE example

ChrisRackauckas:
Currently it StackOverflows in backprop every once in awhile


MikeInnes:
You could try giving this its own manifest and using the `Flux#iterate` branch; that would probably solve the stack overflows.

ChrisRackauckas:
Review? Should work on master now.

MikeInnes:
LGTM, can you just add an entry to the readme as well.

Note that when more recent packages are needed, we have to update the manifest (and check that other models still work on recent versions).

ChrisRackauckas:
Alright cool. Done. @jessebett you might want to run through it.

PR #93, opened 2019-01-29
Implemented A2C on Pendulum Problem

raghav9-97:
Tweaked Tejank's a2c.jl model on Cartpole problem.

### DIFF
--- a/games/pendulum/Manifest.toml
+++ b/games/pendulum/Manifest.toml
@@ -1,0 +1,390 @@
+# This file is machine-generated - editing it directly is not advised
+
+[[AbstractTrees]]
+deps = ["Markdown", "Test"]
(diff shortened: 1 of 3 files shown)


sidhantnagpal:
I think you should retain the authorship of the code. Are there any new changes in this?

raghav9-97:
Yup, I myself wrote the loop implementation and used helper functions and hyperparams from earlier code.

dhairyagandhi96:
Maybe it would be worth looking at convergence further. There are a couple of comments (plotting) still in the script also.

raghav9-97:
I have deliberately commented them out as a user needs to explicitly import Plots in order to render Pendulum environment. I will look out for convergence of the model.

PR #94, opened 2019-01-29
Updated cuda manifests

dhairyagandhi96:
cc @MikeInnes @KristofferC


MikeInnes:
Have you tried the models to verify that they run? Did you manage to reproduce @kristofferC's issue in #83?

MikeInnes:
Whoops, only meant to comment rather than merge for now. I guess that's better than broken API stuff anyway, but if there's any failures here we should definitely find and fix them ASAP.

PR #98, opened 2019-02-03
 Added train-test split. Changed performance metric to RMSE 

shreyas-kowshik:
It would be better to have a train-test split ratio as in the Knet.jl/housing.jl code. 
The performance metric is updated to RMSE now that we have a test set of more than one element.


other/housing/housing.jl
```diff
@@ -16,15 +16,22 @@ isfile("housing.data") ||
...
 
 x = rawdata[1:13,:] |> gpu
 y = rawdata[14:14,:] |> gpu
 
 # Normalise the data
 x = (x .- mean(x, dims = 2)) ./ std(x, dims = 2)
 
-# The model
+# Split into train and test sets
+split_index = convert(Int,floor(size(x,2)*split_ratio))
```
MikeInnes: I think this can be written as `floor(Int, ...)`

MikeInnes:
Seems like a good idea to me, thanks.

shreyas-kowshik:
@MikeInnes updated the file

MikeInnes:
:+1: Thanks!

PR #101, opened 2019-02-05
Implementation of Relax method for Bernoulli random variable

pevnak:
Implementation of relax method for Bernoulli random variable from  the paper Backpropagation through the Void: Optimizing control variates for black-box gradient estimation, Will Grathwohl, Dami Choi, Yuhuai Wu, Geoffrey Roeder, David Duvenaud ,https://arxiv.org/abs/1711.00123

The implementation is a copy-cat from https://github.com/duvenaud/relax

### DIFF
--- a/other/relax/relax.jl
+++ b/other/relax/relax.jl
@@ -1,0 +1,117 @@
+#   Implementation of relax method for Bernoulli random variable from  the paper 
+#   Backpropagation through the Void: Optimizing control variates for black-box gradient estimation,
+#   Will Grathwohl, Dami Choi, Yuhuai Wu, Geoffrey Roeder, David Duvenaud 
+# 
+#   https://arxiv.org/abs/1711.00123
+#
+#   The implementation is a copy-cat from https://github.com/duvenaud/relax
+
+using Flux, Statistics
+using Flux.Tracker
+
+expit(x) = 1 / (1 + exp(-x))
+
+logit(x) = log(x / (1 - x))
+
+heaviside(z) = z >= 0
+
+bernoulli_softmax(z, log_temperature) = expit(z / exp(log_temperature))
+
+logistic_sample(noise, mu = 0, sigma=1) = mu + logit(noise) * sigma
+
+function logaddexp(x, y) 
+    m =  max(x, y)
+    log(exp(x - m) + exp(y - m)) + m
+end
+
+function logistic_logpdf(x, mu = 0, scale = 1)
+    y = (x - mu) / (2 * scale)
+    -2 * logaddexp(y, -y) - log(scale)
+end
+
+bernoulli_sample(logit_theta, noise) = logit(noise) < logit_theta
+
+# relaxed_bernoulli_sample(logit_theta, noise, log_temperature) = bernoulli_softmax(logistic_sample(noise, logit_theta), log_temperature)
+relaxed_bernoulli_sample(logit_theta, noise, log_temperature) = bernoulli_softmax(logistic_sample(noise, expit.(logit_theta)), log_temperature)
+
+"""
+    Computes p(u|b), where b = H(z), z = logit_theta + logit(noise), p(u) = U(0, 1)
+"""
+function conditional_noise(logit_theta, samples, noise)
+    uprime = expit.(-logit_theta)
+    @. samples * (noise * (1 - uprime) + uprime) + (1 - samples) * noise * uprime
+end
+
+"""
+    log Bernoulli(targets | theta), targets are 0 or 1.
+"""
+bernoulli_logprob(logit_theta, targets) = -logaddexp(0, -logit_theta * (targets * 2 - 1))
+
+
+"""
+    function reinforce(θ, u, x)
+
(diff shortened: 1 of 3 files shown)


other/relax/relax.jl
```diff
@@ -0,0 +1,117 @@
...
+ϕ = params(surrogate)
+push!(ϕ, τ)
+nsamples = 10
+
+opt = ADAM(0.1)
+for i in 1:1000
+    u, v = rand(D, nsamples), rand(D, nsamples)
+    fVal = relax_all(θ, ϕ, τ, surrogate, u, v, f)
+    Flux.Optimise.update!(opt, ϕ)
+    Flux.Optimise.update!(opt, [θ])
```
jessebett: This is knit-picky, but is it possible to write `θ` and `ϕ` in such a way that they're interfaced in the same way?

I remember this being quite annoying when I was working with LAX.

E.g. `ϕ` you have `Flux.Optimise.update!(opt, ϕ)` but `θ` needs to be wrapped in an array `Flux.Optimise.update!(opt, [θ])`.

Or above, 

```
    for p in ϕ
        p.grad .= Flux.data(∂var_∂ϕ[p])
    end
``` 

but `θ.grad .= dropdims(mean(Flux.data(∂f_∂θ), dims = 2), dims = 2)`

I don't know if this is worthwhile, but for aesthetics and to avoid confusion, it would be nice to have both of these sets of parameters handled with the same syntax.

pevnak: Frankly speaking, I do not know how to do that at the moment. Relax (and reinforce as well) are weird in the sense that they encapsulated the forward part of the function and the loss as well. Moreover, loss needs to output values on individual samples.

I do not know, how to do this from the software engineering point of view.

pevnak:
I have tried to improve handling coefficients of \theta and \phi.

other/relax/relax_driver.jl
```diff
@@ -0,0 +1,21 @@
...
+include("relax.jl")
+
+D = 100
+θ = param(zeros(D))
+c = collect(range(0,1,length = D))
+f(x) = sum((x .- c).^2, dims = 1)
+b = BernoulliRelax(f, Chain(Dense(D, 5, relu), Dense(5, 1)))
+
+ϕ = params(b)
+push!(ϕ, θ)
```
jessebett: What is this doing? Is it combining the sets of parameters into one? I don't think this is better.

other/relax/relax.jl
```diff
@@ -112,30 +112,27 @@ function BernoulliRelax(f, surrogate)
...
-    mod(i, 10) == 0 && println(i, " fVal = ", mean(fVal))
+"""
+    This is the workhorse function calculating gradient with respect θ and also 
+    calculating surrogate 
+"""
+function ∇relax(θ::Matrix, m::BernoulliRelax,  u = rand(size(θ)), v = rand(size(θ)))
+    func_vals, ∂f_∂θ = relax(param(θ), u, v, m.τ, m.surrogate, m.f)
+    ∂var_∂ϕ = Flux.gradient(() -> sum(∂f_∂θ .^ 2) ./ size(u, 2), Params(m.ϕ); nest = true)
+    
+    foreach(p -> p.grad .= Flux.data(∂var_∂ϕ[p]), m.ϕ)
```
jessebett: So maybe this is unintuitive, because the ϕ gradients are being updated here, but only θ is being given as an argument.

other/relax/relax_driver.jl
```diff
@@ -0,0 +1,21 @@
...
+nsamples = 10
+
+opt = ADAM(0.1)
+@elapsed for i in 1:2000
+    fVal = ∇relax!(θ, b, nsamples)
+    update!(opt, ϕ)
```
jessebett: The confusion above manifests itself here. It looks like you're doing some gradient of relax and it only takes in θ yet in the line below, `update!(opt, ϕ) ` so clearly something happened to the ϕ gradients but it's not obvious.

What I meant before by having the same interface is that the goal is that there should be two lines that look like:
```
update!(opt, θ)
update!(opt, ϕ)
```

pevnak: I do not know, how to really resolve this. May-be, to overload api of Flux.
Something like
```
Flux.gradient(relax,...)
```
Other option I am thinking about would be to try to seamlessly include it into the Flux TrackedArray.

Frankly speaking, I do not know, how to do this at the moment.

MikeInnes:
This could use some readme/walkthrough for someone who isn't familiar with the area. Are there still open Flux API questions? I can try to help with those things if they are narrowed down a bit.

pevnak:
Hi Mike,

I would like to get back to this again. I can write a readme explaining, what the method does. The only mild issue is that I do not know, how to include latex into readme. It would be the best, if I can add some equations.

MikeInnes:
I suggest avoiding latex in the readme but you can add some in the code using jupyter's syntax. That will work if the script is converted to a notebook via literate.jl. The readme can just be an overview with the script giving a more detailed walkthrough.

PR #104, opened 2019-02-07
Improve convolutional MNIST model

staticfloat:
Also add a more complete training harness, demoing how simple it is to
do things like save model weights, early-exit, add learning rate
schedules, etc...

This model achieves 99%+ after ~20 epochs of training.


MikeInnes:
LGTM. Presumably the CUDA manifest will need updating as well (will this work on GPU?).

staticfloat:
Ah, I didn't realize there was a separate CUDA manifest.  Nice.  I'll update that and verify that everything works.

MikeInnes:
Yeah, it's worth giving the other models in the folder a quick run through -- should be fine but just in case.

dhairyagandhi96:
Ref: https://github.com/FluxML/model-zoo/pull/71/commits/312a69fa5442b562f0298864cb20c803ec88a209

This is from #71 
Just in case

MikeInnes:
Ah yes. We should split that version in to the `tutorials` folder and keep them separate.

staticfloat:
Tested on CUDA; sometimes I forget just how much faster hardware accelerators can be.  Confirmed that `conv.jl`, `mlp.jl` and `vae.jl` all work with the new manifest.

staticfloat:
@Roboneet Just a heads up that this is the new model that I want to replace on the website.  I can get you a BSON file later (or, of course, you can just run `conv.jl` locally to generate it) but the model is still quite simple, just a few things changed (channel count, padding, etc...):

```julia
model = Chain(
    # First convolution, operating upon a 28x28 image
    Conv((3, 3), 1=>16, pad=(1,1), relu),
    x -> maxpool(x, (2,2)),

    # Second convolution, operating upon a 14x14 image
    Conv((3, 3), 16=>32, pad=(1,1), relu),
    x -> maxpool(x, (2,2)),

    # Third convolution, operating upon a 7x7 image
    Conv((3, 3), 32=>32, pad=(1,1), relu),
    x -> maxpool(x, (2,2)),

    # Reshape 3d tensor into a 2d one, at this point it should be (3, 3, 32, N)
    # which is where we get the 288 in the `Dense` layer below:
    x -> reshape(x, :, size(x, 4)),
    Dense(288, 10),

    # Finally, softmax to get nice probabilities
    softmax,
)
```

This model should be able to achieve at least 99% accuracy on the test set.

Issue #69, opened 2018-10-24
Next steps for notebooks and tutorials


MikeInnes:
With #59 we have good infrastructure for generating notebooks. There are a couple more steps we need:

* Content. This could be taken from existing open-source tutorials (#68) and we should also port existing notebooks like [these](https://github.com/mikeinnes/flux-notebooks) and [these](https://github.com/aviks/mltraining).
* Auto build notebooks – we can do this on our CI, like documenter, and push them to a branch of this repo.
* JuliaBox – really just the same as the above, but push them to the [JuliaBoxTutorials](https://github.com/JuliaComputing/JuliaBoxTutorials) repo, which gets synced to JuliaBox users' drives.
* Flux Website – we should have a tutorials section on the website, built from these notebooks. We'll need to build HTML from the notebooks we have (ideally with niceties like ToC, but that's secondary) and push them to the Flux website's github pages.

Useful resources:

* [PyTorch tutorials](https://pytorch.org/tutorials/)
* [MXNet – The Straight Dope](https://gluon.mxnet.io)
* Knet [tutorials](https://github.com/denizyuret/Knet.jl/tree/master/tutorial) and [examples](https://github.com/denizyuret/Knet.jl/tree/master/examples)

merckxiaan:
I've tried to implement [PyTorch's seq2seq translation tutorial](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html) (for the model itself I've also looked at [Fast.ai's nmt lesson](https://course.fast.ai/lessons/lesson11.html)).

I've written my process down in [this](https://nbviewer.jupyter.org/github/merckxiaan/flux-seq2seq/blob/master/seq2seq%20in%20flux.ipynb#The-Model) notebook

The model's performance is pretty bad and I'm not sure why that is, but I believe this could perhaps be resolved quickly by someone with more knowledge and experience.

Thanks
Jules

MikeInnes:
That looks great. Any chance you can put it up as a PR in a Literate.jl format? Then we can look at debugging performance.

PR #108, opened 2019-02-26
Seq2seq tutorial

merckxiaan:
As suggested by @MikeInnes in #69, I'm making a pull request with a seq2seq tutorial in literate.jl format.

I'm hoping someone could have a look at this model since performance (translation quality) isn't very good.

Thanks


MikeInnes:
Would be nice to figure out the training issues so that this can get more interesting results before we merge. Also, you shouldn't need to commit the `ipynb` file here since we have a conversion script that will generate it from the latest `.jl` version.

merckxiaan:
Thanks for the feedback, I've removed the notebook version.
As for the translation quality, I'm afraid I'm unable to fix this, although I'm pretty certain I'm making a silly mistake somewhere.

PR #109, opened 2019-02-28
Add logistic regression example on iris dataset.

JoshuaWhittemore:
This is a super simple model aimed at helping beginners get started with Flux.

It depends on an 'iris' Data module in Flux, PR here: https://github.com/FluxML/Flux.jl/pull/652.

If it doesn't make sense to have the iris Data module in Flux I can move that over here.

Please make any suggestions/changes you think might be appropriate.  Thank you!

### DIFF
--- a/other/iris/iris.jl
+++ b/other/iris/iris.jl
@@ -1,0 +1,67 @@
+
+
+using Flux
+using Flux: crossentropy, normalise, onecold, onehotbatch
+using Statistics: mean
+
+
+labels = Flux.Data.Iris.labels()
+features = Flux.Data.Iris.features()
+
+
+# Subract mean, divide by std dev for normed mean of 0 and std dev of 1.
+normed_features = normalise(features, dims=2)
+
+
+klasses = sort(unique(labels))
+onehot_labels = onehotbatch(labels, klasses)
+
+
+# Split into training and test sets, 2/3 for training, 1/3 for test.
+train_indices = [1:3:150 ; 2:3:150]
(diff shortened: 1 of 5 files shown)


other/iris/iris.jl
```diff
@@ -0,0 +1,67 @@
...
+)
+
+loss(x, y) = crossentropy(model(x), y)
+
+# Gradient descent optimiser with learning rate 0.5.
+optimiser = Descent(0.5)
+
+
+# Start Training.
+for epoch in 1:100
```
MikeInnes: This could be done with something like `data = Iterators.repeat((X, Y), 100)`.

joshua-whittemore:
Hi Mike, I've changed the call to `train!` to use an iterator as you suggested.  Let me know if you have any other suggestions.

One thing - this model depends on FluxML/Flux.jl#652 for its dataset.  Thank you!

MikeInnes:
That PR is merged, so presumably this just needs the manifest to be updated?

JoshuaWhittemore:
Yes I think so.  I was going to do that today.  I just need to figure out how to require that it use Flux 0.8.1+ I think.

JoshuaWhittemore:
Err, 0.8.1 or 0.8.1+ I guess.

JoshuaWhittemore:
I've updated the manifest.  Let me know if there's anything else I should do.  Thanks!

MikeInnes:
Thanks!

PR #113, opened 2019-03-07
Implementation of memory networks

deepaksuresh:
Julia/Flux implementation of End to end memory networks for language modelling and question answering


MikeInnes:
This looks generally good but could use a bit more instructions/walkthrough – otherwise I'm not sure how to start playing around with it. Some simple steps to getting it running and showing some neat question answering demos would be great.

PR #115, opened 2019-03-08
Meta-Learning

domluna:
Shortened/simplified my blog post https://www.domluna.me/meta-learning/ into a script suitable for the zoooo.


MikeInnes:
This is very nice, thanks!

It would be useful to open an issue to discuss the need for the `Linear` layer here. Hopefully we can make the builtins more flexible so this kind of thing is less necessary.

PR #117, opened 2019-03-15
Housing Multilayer Percepton Regressor

elifah93:
Adding model using MLP Regressor since I did not find any examples of this model on the web

### DIFF
--- a/other/housing/housing-mlpr.jl
+++ b/other/housing/housing-mlpr.jl
@@ -1,0 +1,29 @@
+using Flux, Statistics, DelimitedFiles
+using Flux: mse, throttle, gpu, @epochs
+using Base.Iterators: repeated
+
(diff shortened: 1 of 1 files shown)


other/housing/housing-mlpr.jl
```diff
@@ -0,0 +1,29 @@
+using Flux, Statistics, DelimitedFiles
+using Flux: mse, throttle, gpu, @epochs
+using Base.Iterators: repeated
+
+cd(@__DIR__)
+
+isfile("housing.data") || download("https://raw.githubusercontent.com/MikeInnes/notebooks/master/housing.data", "housing.data")
```
MikeInnes: Can't you use the same URL as the other file?

MikeInnes:
I like the idea!

PR #127, opened 2019-04-22
Fix x type to match m

ygorcanalli:
Changing `x` type from default `Float64` to `Float32`, to match `Dense` on comparison:

```julia
m = Dense(10, 5)

x = rand(Float32, 10)

m(x)
#-
m(x) == m.W * x .+ m.b
```


dhairyagandhi96:
Thanks for the patch! 👍

PR #134, opened 2019-05-07
Added link to define a confusion matrix since I was confused when I r…

logankilpatrick:
…ead this initially.

added #To avoid confusion, here is the definition of a Confusion Matrix: https://en.wikipedia.org/wiki/Confusion_matrix


dhairyagandhi96:
Thanks!

logankilpatrick:
No problem!

PR #140, opened 2019-06-04
Update 60-minute-blitz.jl

nicolasbarra:
Derivative no longer works, so gradient should be used here
Leads to confusion for beginners attempting to follow tutorial

### DIFF
--- a/tutorials/60-minute-blitz.jl
+++ b/tutorials/60-minute-blitz.jl
@@ -119,9 +119,9 @@ f(5)
 # In simple cases it's pretty easy to work out the gradient by hand – here it's
 # `6x+2`. But it's much easier to make Flux do the work for us!
 
-using Flux.Tracker: derivative
+using Flux.Tracker: gradient
 
-df(x) = derivative(f, x)
+df(x) = gradient(f, x)
 
 df(5)
 
@@ -129,7 +129,7 @@ df(5)
 # as `6x+2`. We can even do this multiple times (but the second derivative is a
 # fairly boring `6`).
 
-ddf(x) = derivative(df, x)
+ddf(x) = gradient(df, x)
 
 ddf(5)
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
Thanks, but note that if you update the syntax you'll likely also need to update the project/manifest files.

nicolasbarra:
> Thanks, but note that if you update the syntax you'll likely also need to update the project/manifest files.

What do you mean by this?

MikeInnes:
See [these docs](https://julialang.github.io/Pkg.jl/dev/environments/). The `.toml` files in this folder provide a set of packages that work with this code. After your changes, the code works with a different (newer) set of packages.

dehann:
Thanks for the fix, I just ran into this same issue on derivative.

nicolasbarra:
> See [these docs](https://julialang.github.io/Pkg.jl/dev/environments/). The `.toml` files in this folder provide a set of packages that work with this code. After your changes, the code works with a different (newer) set of packages.

I doesn't seem like the packages will need to change, this is just a syntax change that happened in Flux and Flux is already a package there.

MikeInnes:
The project records specific package _versions_. So if there's a syntax change in Flux, the project will load the older version with the old syntax, so it'll break if you change (only) the syntax.

dhairyagandhi96:
You probably want to just activate the environment and call `] up` to get the newest versions of the packages in the environment.

nicolasbarra:
I fixed the package/manifest files 
and made another request here:
https://github.com/FluxML/model-zoo/pull/148#issue-293389656

tutorials/60-minute-blitz.jl
```diff
@@ -119,17 +119,17 @@ f(5)
...
 df(5)
 
 # You can try this with a few different inputs to make sure it's really the same
 # as `6x+2`. We can even do this multiple times (but the second derivative is a
 # fairly boring `6`).
 
-ddf(x) = derivative(df, x)
+ddf(x) = gradient(df, x)
 
 ddf(5)
```
mbeltagy: This does not work on Julia 1.10 and Flux 0.8.3 
See also 
https://github.com/FluxML/Zygote.jl/issues/278

hossein-pourbozorg:
The changes exist in #143, so should we close this PR?

aminya:
@dhairyagandhi96 this can be closed as it fixed by #143

dhairyagandhi96:
Closing since #143 is merged

PR #147, opened 2019-06-14
Another mistake in KL-divergence equation vae.jl

kianaf:
As mentioned by @AbinavRavi in KL-divergence equation we should have 2*logσ instead of logσ.^2 but the other mistake is the sign of this which should be -2*logσ


willtebbutt:
You're totally right, not sure how this slipped through. Many thanks for sorting it out.

MikeInnes:
Thanks both!

PR #143, opened 2019-06-07
Updated models in the zoo to use the latest versions of various packages 

SudhanshuAgrawal27:
Updated the Manifests of cppn, mnist, and cifar10 

Ensured that the sample output for cppn saves to the same directory as the program file.

Ensured that the model created by mnist\conv saves to the same directory as the program file.

Replaced the given sample output of cppn with a new one. 

Tried changing the optimiser from ADAM to Momentum in the cifar10 program but got a better accuracy with ADAM so reverted to it.

PR #148, opened 2019-07-01
Updating package/manifest files to support move from derivative to gradient

nicolasbarra:
Updated the files specified to support this move
https://github.com/FluxML/model-zoo/pull/140#issue-285177436
Derivative no longer works, so gradient should be used here
Leads to confusion for beginners attempting to follow tutorial


MikeInnes:
It would be great to combine this with #140. Otherwise it's much harder for someone to test this code change.

hossein-pourbozorg:
Most of this changes appeared in #143. 
IMO, It's better to work on #143 because of it has progressed more.

dhairyagandhi96:
Closing in favour of #143 
Thanks all!

PR #150, opened 2019-07-09
Update models for maxpool

hossein-pourbozorg:
I just replaced every
```julia
x -> maxpool(x, (N,N))
```
with
```julia
MaxPool((N, N))
```

Before I do this changes, I got
```julia
ERROR: LoadError: MethodError: no method matching maxpool(::TrackedArray{…,Array{Float32,4}}, ::Tuple{Int64,Int64})
```
when I attempted to run
```julia
vision/mnist/conv.jl
vision/cifar10/cifar10.jl
```
examples. And after that, the error disappeared.

PR #156, opened 2019-08-04
maxpool -> MaxPool

kraftpunk97:
`x -> maxpool(x, (2, 2))` errors out. Proposing change to the `MaxPool` layer.


kraftpunk97:
closing because a similar PR with more progress is already in  queue #150

PR #171, opened 2019-09-30
Changed to zygote

UmashankarTriforce:
These files had `Flux.Zygote` as import mechanism, which caused `Unknown variable Zygote` error. This PR fixes that by changing to `Using Zygote`


dhairyagandhi96:
The Project in that branch uses Flux#master till Flux is tagged. This doesn't seem necessary.

UmashankarTriforce:
Okay I'll close this

PR #172, opened 2019-10-01
Removing unused keyword arg in char RNN example

mcognetta:
Also removing an old comment.


dhairyagandhi96:
Perfect, thanks for the patch @mcognetta !

PR #173, opened 2019-10-01
Move to a unified environment and update to Flux @0.9

dhairyagandhi96:
What this does:
- With Flux 0.9, we don't need to hold CuArrays as an external dependency, use this to simplify the model zoo
- Rather than have multiple environments, per model, have a single environment that is available to all models
- Makes it easy to maintain and add models

What remains:
- [ ] Check everything on the GPU - housing/housing and vision/mnist/autoencoder complain (they work with master)
- [x] Metalhead needs updating (required for cifar, 60 min blitz) - done, needs tagging
- [x] Remove all `cuda` folders
- [x] Document usage of the model-zoo post this change

Retroactively also solves #170


MikeInnes:
> Rather than have multiple environments, per model, have a single environment that is available to all models

I think it's good to share environments where possible, especially for a well-maintained core subset of models. But we probably need to have manifests too for some models, either just for total reproducibility or because they have heavier dependencies (e.g. plots or data loaders).

dhairyagandhi96:
Right, some models like dP and some diffeq ones haven't been touched, for this exact reason. I am probably going to keep them intact. We should message this properly, with a disclaimer somewhere maybe?

MikeInnes:
Yeah, perhaps a `contrib` folder with miscellaneous contributed models, and the rest we commit to keeping up to date.

dhairyagandhi96:
Makes sense, those should also just be updated for the since some of the APIs have moved on since

dhairyagandhi96:
Good to get some eyes on this about now.

Notebooks: https://github.com/dhairyagandhi96/model-zoo/tree/notebooks/notebooks

MikeInnes:
I like the new layout; it's nice to have a small set of "blessed" models that we maintain, and should always work on latest release, and then the contrib models can have specific manifests and more instructions on how to instantiate etc.

We of course need to update to tagged releases when they are out, but otherwise you should just merge once you are happy with it.

dhairyagandhi96:
Should merge here now, to make it easier to make merging the Zygote branch easier.

PR #174, opened 2019-10-07
Update conv.jl to use MaxPool

nlw0:
Updating code to use the new MaxPool class. Does not work with the latest Flux otherwise.


hossein-pourbozorg:
Related PR #150

nlw0:
#150 is where it's at

PR #177, opened 2019-11-21
Add GitLab config to get bot up

dhairyagandhi96:
Useful to get model zoo runs on the GPU with GitLab. Steps in the direction to be able to trigger runs using a comment on Flux.jl to build specified models

cc @maleadt


dhairyagandhi96:
Requires #173

dhairyagandhi96:
Merging this now, so I can iterate on it as needed, otherwise it's a bit slow to know how GitLab will handle this.

PR #184, opened 2019-12-19
Update README.md

tpoisot:
The link to the *MetaLearning.jl* example was wrong - I fixed it.


dhairyagandhi96:
Looks good , and thanks for catching that!

Would you be up for fixing the links for all the models that have moved to `contrib` recently?

tpoisot:
I hadn't noticed the other models! Absolutely, let me do this in the next five minutes.

dhairyagandhi96:
These look great! Thanks for catching these @tpoisot !

PR #187, opened 2020-01-03
Make vision models run on Flux v0.10.0 with GPU

tanhevg:


CarloLucibello:
I had a look at this PR and run the  mnist/mlp.jl script, everything looks and works fine. 
@dhairyagandhi96  can we merge this PR right now? 
it's really a shame the examples in model-zoo have not been working on the current Flux version for many months, this is making life very complicated for new users.

FerreolS:
> I had a look at this PR and run the mnist/mlp.jl script, everything looks and works fine.
> @dhairyagandhi96 can we merge this PR right now?
> it's really a shame the examples in model-zoo have not been working on the current Flux version for many months, this is making life very complicated for new users.

I'agree, as model-zoo is presented as an entrypoint in Flux for newcomers
Fix the problem also fixed in #197

dhairyagandhi96:
Do we really need to have adjoints and such defined? In the zygote branch we didn't need to do that.

tanhevg:
@dhairyagandhi96 , custom adjoint for `augment` is needed because Zygote could not differentiate through call to `gpu` at the time when the PR was created, as pointed out by the comment.

dhairyagandhi96:
I think that's been resolved (that was the reason I'd waited on updating then)

Good to do a quick check

tanhevg:
Could you point me to the PR that resolved the `gpu` differentiation issue, please?

dhairyagandhi96:
https://github.com/FluxML/Zygote.jl/pull/423

tanhevg:
Indeed, the custom adjoint is not required any more, I have removed it.

tanhevg:
@dhairyagandhi96 , plese review/merge

CarloLucibello:
@dhairyagandhi96 bump

vision/mnist/conv.jl
```diff
@@ -91,6 +101,11 @@ for epoch_idx in 1:100
     global best_acc, last_improvement
     # Train for a single epoch
     Flux.train!(loss, params(model), train_set, opt)
+    
+    if anynan(paramvec(model))
+        @error "NaN params"
+        break
+    end
```
CarloLucibello: I would remove this check, it's not needed in an intro example and also is not something one commonly does in practice

CarloLucibello: If we want to keep it, let's make the check more efficient:
```julia
anynan(m) = any(any(isnan, p) for p in params(m))
```

dhairyagandhi96: I think I'm in favor of removing it too

vision/mnist/autoencoder.jl
```diff
@@ -1,8 +1,12 @@
 using Flux, Flux.Data.MNIST
 using Flux: @epochs, onehotbatch, mse, throttle
 using Base.Iterators: partition
-using Juno: @progress
-# using CuArrays
+using CUDAapi
+if has_cuda()
+    @info "CUDA is on"
```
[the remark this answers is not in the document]
tanhevg: Flux does not "message CUDA functionality" at the moment, so I'd rather keep it for now. The main purpose of using CUDAapi is to be able to import CuArrays, and turn off scalar indexing on  GPU-enabled machines.

vision/mnist/conv.jl
```diff
@@ -68,17 +74,21 @@ model = gpu(model)
 # Make sure our model is nicely precompiled before starting our training loop
 model(train_set[1][1])
 
+# We augment `x` a little bit here, adding in random noise. 
+augment(x) = x .+ gpu(0.1f0*randn(eltype(x), size(x)))
+
+paramvec(m) = vcat(map(p->p[:], params(m))...)
+anynan(x::AbstractArray{<:AbstractFloat}) = any(isnan.(x))
```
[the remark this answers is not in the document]
tanhevg: Done

Project.toml
```diff
@@ -1,8 +1,10 @@
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
+CUDAapi = "3895d2a7-ec45-59b8-82bb-cfc6a382f9b3"
 Cascadia = "54eefc05-d75b-58de-a785-1a3403f0919f"
 Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
+CuArrays = "3a865a2d-5b23-5a0f-bc46-62713ec82fae"
```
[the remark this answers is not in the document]
tanhevg: Removed Zygote; see my comment above why we need CuArrays - essentially, to turn off scalar indexing.

vision/mnist/conv.jl
tanhevg: I get this error from time to time on my machine, when some of the parameters turn `NaN` and the accuracy then starts taking silly values and the model never converges. It could be related to some peculiar combination of CUDNN and GPU drivers that I have; would need to dig around a bit more to post a sensible bug report. I'd like to keep this check for now, to avoid saving models with NaN parameters.

vision/mnist/conv.jl
```diff
@@ -68,17 +74,21 @@ model = gpu(model)
 # Make sure our model is nicely precompiled before starting our training loop
 model(train_set[1][1])
 
+# We augment `x` a little bit here, adding in random noise. 
+augment(x) = x .+ gpu(0.1f0*randn(eltype(x), size(x)))
```
CarloLucibello: we should also remove this kind of data augmentation, it's not common at all, I don't see why it should get into one of the basic scripts that people look at as a template for their code.

vision/mnist/conv.jl
CarloLucibello: I understand but the purpose of this script is to serve as a basic template, so we should have here only standard stuff

Project.toml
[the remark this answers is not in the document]
dhairyagandhi96: Using Flux.CuArrays and Flux.Zygote achieve the same

vision/mnist/conv.jl
dhairyagandhi96: A message about basic augmentation would be nice, just to show how to get it working

vision/mnist/autoencoder.jl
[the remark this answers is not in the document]
dhairyagandhi96:
```Julia
julia> using Flux
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
```

Should be enough?

CarloLucibello:
bump, we're almost done here

CarloLucibello:
let's merge this as it is, it is definitely an improvement. In any case, we will have to revisit later all of these scripts to make them look like real-world scripts

PR #188, opened 2020-01-03
Judge a book by its cover cnn

Akshat-mehrotra:
Added Code for the judge a book by its cover task.


Akshat-mehrotra:
current accuracy of 21.3%, similar to what is described in the paper.

PR #189, opened 2020-01-04
Annotation for Character RNN 

chinglamchoi:


chinglamchoi:
Thank you!!

text/char-rnn/char-rnn.jl
```diff
@@ -1,59 +1,80 @@
...
+
 using Flux
 using Flux: onehot, chunk, batchseq, throttle, crossentropy
 using StatsBase: wsample
 using Base.Iterators: partition
 
+# # 2. Data Download & Pre-processing
+# - Source of data: Shakespeare text from https://cs.stanford.edu/people/karpathy/char-rnn/
+# - Generate character tokens
+# - Partition in batches for input
```
dhairyagandhi96: Probably good to talk about how these batches are divided. The comments in the code blocks would do well as Markdown cells that detail the why and how in a couple lines

chinglamchoi: Sure! I'll add these details in a few days' time if you don't mind, it's actually testing/examination period for my school!

PR #198, opened 2020-02-05
mlp with gpu,minibatch

mrchaos:
fix loss NaN
use minibatch


CarloLucibello:
this is too much stuff for a single PR, should split into several ones, e.g. one for each script. Also, files contain lot of comments in korean.

mrchaos:
I've kept the sources simple

PR #203, opened 2020-02-14
How about we don't use actively terrible julia code as a common demo?

oxinabox:
I fixed it from being the worst by putting it in a function.
Though I wonder if better would have been to fix it with `const` + making `loss(m) = (x,y)->...`  etc.

### DIFF
--- a/vision/mnist/mlp.jl
+++ b/vision/mnist/mlp.jl
@@ -4,34 +4,32 @@ using Base.Iterators: repeated
 # using CuArrays
 
 # Classify MNIST digits with a simple multi-layer-perceptron
+function demo(
+    imgs = MNIST.images(),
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Ha, not much of a demonstration now is it 😂

tk3369:
I think it's best to promote good practice.  I imagine that people would take a template from model zoo and start manipulating it into their own use cases.

oxinabox:
People do.
E.g, i see very similar code is now in:
https://github.com/PhilipVinc/TensorBoardLogger.jl/pull/60

jamblejoe:
This was suggested for the examples in Flux.jl as well, see https://github.com/FluxML/Flux.jl/pull/1085. 
I find it easiest to suround critical parts in `let ... end` blocks.

PR #207, opened 2020-02-29
DCGAN model for Flux v0.10

matsueushi:
This is a DCGAN implementation for Flux v0.10. I know there are already pending pull requests for DCGAN (https://github.com/FluxML/model-zoo/pull/47, https://github.com/FluxML/model-zoo/pull/111), but they are incompatible with Zygote.

A linear layer is used as the last layer of the discriminator and losses are calculated using `logitbinarycrossentropy`. This is because the combination of `sigmoid` and `binarycrossentropy` may cause numerical issues (https://github.com/FluxML/Flux.jl/issues/914).

It ouputs generated digits for a fixed noise every 1000 iterations. I believe it is helpful to trace its training process :)

0 steps
dcgan_steps_000000

3000 steps
dcgan_steps_003000

6000 steps
dcgan_steps_006000

final result (=9380 steps)
dcgan_steps_009380


vision/mnist/dcgan.jl
```diff
@@ -0,0 +1,126 @@
...
+            train_steps += 1
+        end
+    end
+
+    output_image = create_output_image(gen, fixed_noise)
+    save(@sprintf("dcgan_steps_%06d.png", train_steps), output_image)
+end
+
+cd(@__DIR__)
+train()
```
CarloLucibello: newline here

CarloLucibello:
very nice. Is this the architecture from the original paper (or any other)?. If so, can you add a reference in a comment?

CarloLucibello:
you should add the model to the README

CarloLucibello:
Maybe this is the right moment to start reorganizing this repo. We could have a folder structure like this `domain/model_dataset/`, and each example should come with a Project.toml and a Manifest.toml. Is this too cumbersome? @matsueushi what do you think?

vision/mnist/dcgan.jl
```diff
@@ -0,0 +1,126 @@
...
+        for batch in data
+            # Update discriminator and generator
+            loss_dscr = train_discriminator!(gen, dscr, batch, opt_dscr)
+            loss_gen = train_generator!(gen, dscr, batch, opt_gen)
+
+            if train_steps % VERBOSE_FREQ == 0
+                @info("Train step $(train_steps), Discriminator loss = $(loss_dscr), Generator loss = $(loss_gen)")
+                # Save generated fake image
+                output_image = create_output_image(gen, fixed_noise)
+                save(@sprintf("dcgan_steps_%06d.png", train_steps), output_image)
```
CarloLucibello: you could also add to this PR those figures you produced, it's nice to have a reference

CarloLucibello:
Could you also use https://github.com/JuliaML/MLDatasets.jl instead of Flux.Data.MNIST? It's more future proof since we are going to excise Julia.Data soon

matsueushi:
@CarloLucibello , thank you for reviewing my PR.

> very nice. Is this the architecture from the original paper (or any other)?. If so, can you add a reference in a comment?

Basically the architecture follows the DCGAN tutorial for tensorflow (https://www.tensorflow.org/tutorials/generative/dcgan). I would change the hyperparams if necessary.

> Maybe this is the right moment to start reorganizing this repo. We could have a folder structure like this `domain/model_dataset/`, and each example should come with a Project.toml and a Manifest.toml.
> So, this should go under `vision/dcgan_mnist/`.
> Is this too cumbersome? @matsueushi what do you think?

I'm on board.  It will be easier to maintain models and update deps.

matsueushi:
I will replace Flux.Data.MNIST by MLDatasets

vision/mnist/dcgan.jl
CarloLucibello: sorry, I meant a newline at the very end of the file, so that we don't have github complaining with a red arrow

vision/dcgan_mnist/dcgan.jl
```diff
@@ -0,0 +1,128 @@
...
+using Flux
+using Flux.Data.MNIST
+using Flux.Optimise: update!
+using Flux: logitbinarycrossentropy
+using Images
+using Statistics
+using Printf
+
+const BATCH_SIZE = 128
+const NOISE_DIM = 100
```
CarloLucibello: maybe LATENT_DIM sounds better here

CarloLucibello:
I very much like the overall style, this would be a nice template for the other models. 
My only perplexity is on the use of global variables: I have a REPL based workflow and typically have all parameters as keyword arguments to the main function, then pass them around to the various methods, usually grouping them together in a `Params` struct I create with https://github.com/mauro3/Parameters.jl. Of course, this is just my personal taste, what you did
here is perfectly fine, just wondering what would be the "julian" way to write deep learning scripts that most people would consider as convenient templates for their code

vision/mnist/dcgan.jl
matsueushi: Oh, sorry. I'll fix that

matsueushi:
I used global variables just because they are used in other vision models. So if it is time to change, I switch them to `Params`.

matsueushi:
`Params` is already used by Zygote (https://github.com/FluxML/Zygote.jl/blob/6f17f8b5c40e48e6d7732afb91dba9a1ddac145b/src/compiler/interface.jl#L53-L57) so I used `HyperParams` for struct

vision/dcgan_mnist/dcgan.jl
```diff
@@ -0,0 +1,130 @@
...
+        generator_loss(dscr(gen(noise)))
+    end
+    grad = back(1f0)
+    update!(opt_gen, ps, grad)
+    return loss
+end
+
+function train()
+    # Model Parameters
+    hparams = HyperParams()
```
CarloLucibello: here we can do:
```julia
function train(; kws...)
    hparams = HyperParams(; kws...)
    ....
```

matsueushi: Thanks!

vision/dcgan_mnist/dcgan.jl
```diff
@@ -0,0 +1,130 @@
...
+
+    # Load MNIST dataset
+    images, _ = MLDatasets.MNIST.traindata(Float32)
+    # Normalize to [-1, 1] and convert it to WHCN
+    image_tensor = permutedims(reshape(@.(2f0 * images - 1f0), 28, 28, 1, :), (2, 1, 3, 4))
```
CarloLucibello: ouch, the permutedims is a bit annoying but I guess we cannot do much about it without breaking changes in MLDatasets

CarloLucibello: good catch anyways

CarloLucibello:
could you commit the pngs as well? then I think we are good to go

vision/dcgan_mnist/dcgan.jl
```diff
@@ -0,0 +1,130 @@
+using Base.Iterators: partition
+using Flux
+using Flux.Optimise: update!
+using Flux: logitbinarycrossentropy
+using Images
+using MLDatasets
+using Statistics
+using Parameters
```
CarloLucibello:
```suggestion
using Parameters: @with_kw
```

matsueushi:
> could you commit the pngs as well? then I think we are good to go

Sure, I will rerun a whole training process again to make sure everything works fine and add ouput images

matsueushi:
0 steps
dcgan_steps_000000

3000 steps
dcgan_steps_003000

6000 steps
dcgan_steps_006000

9380 steps
dcgan_steps_009380

vision/dcgan_mnist/dcgan.jl
```diff
@@ -0,0 +1,130 @@
...
+function discriminator_loss(real_output, fake_output)
+    real_loss = mean(logitbinarycrossentropy.(real_output, 1f0))
+    fake_loss = mean(logitbinarycrossentropy.(fake_output, 0f0))
+    return real_loss + fake_loss
+end
+
+generator_loss(fake_output) = mean(logitbinarycrossentropy.(fake_output, 1f0))
+
+function train_discriminator!(gen, dscr, batch, opt_dscr, hparams)
+    noise = randn(Float32, hparams.latent_dim, hparams.batch_size) |> gpu
```
CarloLucibello: this is not ideal, we should avoid data transfer and use `CuArrays.randn` if training on gpu. I don't know what would be an elegant way to do that, maybe we can revisit in the future

matsueushi: we might write something like
```julia
randn!(similar(batch, (hparams.latent_dim, hparams.batch_size)))
```
https://docs.julialang.org/en/v1/stdlib/Random/#Random.randn!, https://github.com/JuliaGPU/CuArrays.jl/blob/da2389f0db46d611a5b625c2ce6bba79aa20163f/src/rand/random.jl#L88-L95

vision/dcgan_mnist/dcgan.jl
```diff
@@ -0,0 +1,130 @@
...
+    opt_gen = ADAM(hparams.lr_gen)
+
+    # Training
+    train_steps = 0
+    for ep in 1:hparams.epochs
+        @info "Epoch $ep"
+        for batch in data
+            # Update discriminator and generator
+            loss_dscr = train_discriminator!(gen, dscr, batch, opt_dscr, hparams)
+            loss_gen = train_generator!(gen, dscr, batch, opt_gen, hparams)
```
CarloLucibello: here part of the computation done in train_discriminator! could be reused in train_generator!, i.e. in principle one need only a single forward of the generator.  

This is also done here https://pytorch.org/tutorials/beginner/dcgan_faces_tutorial.html

I don't know how to do it properly with zygote, needs some thinking. If you have a simple solution in mind  you could add this optimization, otherwise we can revisit in the future

matsueushi: Unfortunately I don't know how to do it, too.

vision/dcgan_mnist/dcgan.jl
CarloLucibello: nice!

CarloLucibello:
better move the images to a subfolder

CarloLucibello:
last tiny detail, consider renaming `batch` to `x`

matsueushi:
Thanks, I applied the changes

CarloLucibello:
terrific works, thanks!

matsueushi:
@CarloLucibello, thank you for your awesome detailed review!

PR #210, opened 2020-03-04
Updated model script formats

AdarshKumar712:
I have updated the script formats of the following models:
1. vision/mnist/mlp.jl
2. other/iris/iris.jl

I have also added Project.toml and Manifest.toml files for iris dataset. Will add Project.toml and Manifest.toml files for the mnist models as well, once there script format are updated. I am currently working on them. I will add them as well pretty soon. Please review these updates, and suggest further changes. 

Update:
I have updated following models as well:
1. other/housing/housing.jl
2. vision/mnist/conv.jl
3. vision/mnist/vae.jl
4. vision/mnist/autoencoders.jl
5. vision/mnist/cifar10/cifar10.jl
6. vision/mnist/cppn/cppn.jl
8. vision/dcgan_mnist/dcgan.jl
7. other/fizzbuzz/fizzbuzz.jl
8. other/bitstring-parity/xor1.jl, xor2.jl, xor3.jl
9. text/char-rnn/char-rnn.jl
10. text/lang-detection/model.jl
11. text/phonemes/0-data.jl, 1-model.jl
12. text/treebank/data.jl, recursive.jl


dhairyagandhi96:
We should avoid adding environments per model, it's one thing that made the zoo very difficult to maintain

AdarshKumar712:
Ok, then I will remove local environment files from here, and from some other places as well, and instead will Update the global environment file.

CarloLucibello:
> We should avoid adding environments per model, it's one thing that made the zoo very difficult to maintain

Do you have a reference for this discussion? I would say quite the opposite: a general manifest for all of the examples in unmaintainable. No one will ever update all the scripts at the same time, so some of them will rot and some will go out of sync with the manifest. Kind of the current situation

AdarshKumar712:
I have a doubt. In the Conv.jl, we are saving the parameters as:
```BSON.@save joinpath(savepath, "mnist_conv.bson") params=cpu.(params(model)) epoch_idx acc```
How one is going reload the model using this way? Shouldn't we add how models can be reloaded as well?
@dhairyagandhi96 @CarloLucibello

File: other/iris/iris.jl [hunk 1]
```diff
@@ -1,68 +1,91 @@
...
+	train_data, test_data = get_processed_data(args)
+
+	# Declare model taking 4 features as inputs and outputting 3 probabiltiies, 
+	# one for each species of iris.
+	model = Chain(
+    		 Dense(4, 3),
+    		 softmax)
+	
+	# Defining loss function to be used in training
+	loss(x, y) = crossentropy(model(x), y)
```
CarloLucibello: for numerical stability use `logitcrossentropy`, remove `softmax` from the model

CarloLucibello:
> I have a doubt. In the Conv.jl, we are saving the parameters as:
> `BSON.@save joinpath(dirname(@__DIR__), "mnist_conv.bson") params=cpu.(params(model)) epoch_idx acc`
> How one is going reload the model using this way? Shouldn't we add how models can be reloaded as well?
> @dhairyagandhi96 @CarloLucibello

you can use `load_params!`. It won't work correctly for models with buffers such as batch norm layers, since won't be saved and reloaded, but for a simple conv net it's fine

other/iris/iris.jl
AdarshKumar712: How here using softmax, crossentropy can be numerically unstable?

AdarshKumar712: Ok, got it. I have updated that.

AdarshKumar712:
@dhairyagandhi96 @CarloLucibello  I have updated few other models as well. Please have a look, and suggest changes wherever required.

dhairyagandhi96:
ref #173

dhairyagandhi96:
The general manifest is expected to work with patch updates, else its a bug, and a major update would largely require a rewrite anyway

CarloLucibello:
> The general manifest is expected to work with patch updates, else its a bug, and a major update would largely require a rewrite anyway

problem is that after a major update no one is going to undertake the update of the whole repo (we have just seen this). What we can hope for (and what we consistently got in the past months) are pointwise PRs updating single scripts. This is why we should just have one manifest per example to have both manageability and consistency

dhairyagandhi96:
We had that earlier and it lead to a case where we missed the guarantee that the models would update accordingly. Maintaining multiple environments was also error prone. This time the reason was that we waited for some Zygote bugs to be resolved, a large change which took time to stabilize

AdarshKumar712:
Sorry I had been busy with my GSoC Proposal, because of which I wasn't able to work on this PR. Its almost done now, I will soon update format of other files as well. 
Btw, @dhairyagandhi96 how about having individual Manifest and Project files for each section? Like Vision, Text, Others. Manifest.toml, Project.toml files for each one. This way it would be easier to maintain each section separately.

AdarshKumar712:
@CarloLucibello @dhairyagandhi96 I have updated all models in text, other and vision section, except for one `flux-next` as I was not clear about its purpose. Please do have a look at the changes and suggest improvements.

jamblejoe:
Is the device keyword of the Args struct in mlp.jl ever used? Is the model actually trained on the gpu?

AdarshKumar712:
Sorry @jamblejoe, I haven't checked for the GPU support for all the scripts. This device keyword was left unused by mistake. I will update this shortly. Thanks for reminding! :)

dhairyagandhi96:
What remains to be done here?

AdarshKumar712:
@dhairyagandhi96 I have updated all the models except `others/Flux-next` and `contrib` section. All models work with CPU. But the text models throws an error with the  GPU because of `Onehotmatrix`. Other than that vision models work fine with GPU.

Also for `char-rnn.jl`, the model performs better without reset command there. So I haven't included that in `char-rnn.jl` .

vision/mnist/mlp.jl
```diff
@@ -1,43 +1,89 @@
...
+end
 
-labels = MNIST.labels()
-# One-hot-encode the labels
-Y = onehotbatch(labels, 0:9) |> gpu
+function model(; imgsize=(28,28,1), nclasses=10)
+    return Chain(
+ 	    Dense(prod(imgsize), 32, relu),
+            Dense(32, nclasses),
+  	    softmax)
```
CarloLucibello, line 44: could you remove `softmax` for all models and use `logitcrossentropy` as a loss instead? this should be numerically more stable

text/char-rnn/char-rnn.jl
```diff
@@ -51,10 +50,9 @@ function train(; kws...)
 
     # Constructing Model
     m = Construct_model(N)
-    m = gpu(m)
 
     function loss(xs, ys)
-      l = sum(crossentropy.(m.(gpu.(xs)), gpu.(ys)))
+      l = sum(logitcrossentropy.(m.(xs), ys))
```
CarloLucibello, line 55: is it ok to have a broadcast here?

text/phonemes/1-model.jl
```diff
@@ -79,7 +79,7 @@ function train(; kws...)
     @info("Constructing Model...")
     state, encode = Construct_model(args)
 
-    loss(x, yo, y) = sum(crossentropy.(model(x, yo, state, encode), y))
+    loss(x, yo, y) = sum(logitcrossentropy.(model(x, yo, state, encode), y))
```
CarloLucibello, line 82: also here, do we want to broadcast?

text/char-rnn/char-rnn.jl
AdarshKumar712: Yes, here without the broadcast, it throws the following error:
```
MethodError: no method matching isless(::Array{Float32,2}, ::Array{Float32,2})
Closest candidates are:
  isless(!Matched::Missing, ::Any) at missing.jl:87
  isless(::Any, !Matched::Missing) at missing.jl:88
```
as here `xs` is an array of 50 elements where each element is a 2D array (N x 50) which is input for the model. However I think here it would be better to have `mean` rather than `sum`

text/phonemes/1-model.jl
AdarshKumar712: This also had the same problem as the above.

CarloLucibello:
ok, let's rebase and merge this

text/char-rnn/char-rnn.jl
```diff
@@ -1,59 +1,84 @@
...
+    return Xs, Ys, N, alphabet
+end
 
-m = Chain(
-  LSTM(N, 128),
-  LSTM(128, 128),
-  Dense(128, N),
-  softmax)
+# Function to construct model
+function Construct_model(N)
```
CarloLucibello, line 38: this naming is unjulian, use `construct_model` or `buid_model`

vision/mnist/mlp.jl
```diff
@@ -1,43 +1,88 @@
...
+end
+
+function get_data(args)
+    # Loading Dataset	
+    xtrain, ytrain = MLDatasets.MNIST.traindata(Float32)
+    xtest, ytest = MLDatasets.MNIST.testdata(Float32)
+	
+    # Reshape Data for flatten the each image into linear array
+    xtrain = reshape(xtrain,:,size(xtrain,3))
+    xtest = reshape(xtest,:,size(xtest,3))
```
CarloLucibello, line 28: let's show `Flux.flatten` here

vision/mnist/conv.jl
```diff
@@ -68,10 +68,9 @@ function Construct_model(args; imgsize = (28,28,1), nclasses = 10)
     Conv((3, 3), 32=>32, pad=(1,1), relu),
     MaxPool((2,2)),
 
-    # Reshape 3d tensor into a 2d one, at this point it should be (3, 3, 32, N)
-    # which is where we get the 288 in the `Dense` layer below:
-    x -> reshape(x, :, size(x, 4)),
-    Dense(prod(CNN_output_size), 10))
+    # Reshape 3d tensor into a 2d one using `Flux.flatten`, at this point it should be (3, 3, 32, N)
+    x -> Flux.flatten(x),
```
CarloLucibello, line 72:
```suggestion
    flatten
```

CarloLucibello:
thanks for all this work!

MikeInnes:
See https://github.com/FluxML/model-zoo/pull/226#issuecomment-627361444, but also this PR did not update the project or manifest files, so the code does not have a known-good version.

AdarshKumar712:
The project file was not updated earlier because it was unclear how were we going to handle project dependencies. Though Dhairya earlier pointed out([comment](https://github.com/FluxML/model-zoo/pull/210#issuecomment-598696204)) that having many project files makes it harder to handle, however at the same time having a single project file makes it tough at the time of making updates([comment](https://github.com/FluxML/model-zoo/pull/210#issuecomment-598694993)). To this I suggested that it would be better if we have a section-wise Project and Manifest file like for Vision, Text, Other. However, this discussion didn't go to a conclusion at that time, and remain undecided. 
I hope it won't take much time to update the project files once it's decided on how we are going to keep the project files.

MikeInnes:
Having code in the zoo that needs a project file but doesn't have one definitely seems like the worst of all worlds.

AdarshKumar712:
I have added the Project files, #232

PR #217, opened 2020-03-19
change `truncate!` to `reset!` in char-rnn

mcognetta:
The method `truncate!` is no longer used and `reset!` is now preferred. Changing it so that this example runs without errors.


erlebach:
Truncate! Is still used in the master branch. But your change was supposedly merged with the master branch 23 days ago. What happened?

dhairyagandhi96:
Thanks for the patch! Apologies for the delay

PR #219, opened 2020-03-27
Fix the hyperlink for the games dir in README

vincentzhang:
The "game" directory has been moved into the contrib directory and hence the url in README needs an update.


dhairyagandhi96:
Thanks!

PR #226, opened 2020-04-25
Refactoring VAE model

matsueushi:
This pull request is intend to update VAE model to follow the modern style used by lenet_mnist and dcgan_mnist. 

* Progress bars
* TensorBoard logging
* Saving model as a BSON
* Visualization of original/reconstructed image

Original
original

5 epochs
epoch_5

10 epochs
epoch_10

20 epochs
epoch_20

I also added a recipe (vae_plot.jl) for
* Loading model
* Scatter plot of training samples
* Visualization of the latent manifold.

clustering

manifold


vision/vae_mnist/vae_mnist.jl
```diff
@@ -0,0 +1,173 @@
...
+using Parameters: @with_kw
+using ProgressMeter: Progress, next!
+using TensorBoardLogger: TBLogger, tb_overwrite
+using Random
+
+# load MNIST images and return loader
+function get_data(batch_size)
+    xtrain, ytrain = MLDatasets.MNIST.traindata(Float32)
+    # MLDatasets uses HWCN format, Flux works with WHCN 
+    xtrain = reshape(permutedims(xtrain, (2, 1, 3)), 28^2, :)
```
CarloLucibello, line 26: it turns out that MLDatasets already returns in WHCN so no need for `permutedims` here.
Images instead works with HWCN, so the permutedims should go into `convert_to_image`. 
Could you fix this also for the DCGAN and LeNet scripts?

vision/vae_mnist/vae_mnist.jl
```diff
@@ -0,0 +1,173 @@
...
+    μ, logσ, decoder(z)
+end
+
+function model_loss(encoder, decoder, λ, x, device)
+    μ, logσ, decoder_z = reconstuct(encoder, decoder, x, device)
+    len = size(x)[end]
+    # KL-divergence
+    kl_q_p = 0.5f0 * sum(@. (exp(2f0 * logσ) + μ^2 -1f0 - 2f0 * logσ)) / len
+
+    logp_x_z = -sum(binarycrossentropy.(decoder_z, x)) / len
```
CarloLucibello, line 63: logitbinarycrossentropy here?

CarloLucibello: sorry, never mind, you have a simgoid as the last activation. If you experience some  nnumerical instabilities, you could try with removing the sigmoid and using logitbinarycorssentropy instead

vision/vae_mnist/vae_mnist.jl
matsueushi: Thanks, I removed `permutedims` from `get_data`.

vision/vae_mnist/vae_mnist.jl
matsueushi: I switched sigmoid + bce to logitbce and inserted sigmoid to `convert_to_image`.

CarloLucibello:
thank you, this is quite useful

MikeInnes:
I'm not sure that all these refactorings are an improvement. The VAE example use to be very concise and clearly show the mathematical intent. Now that's obscured by a bunch of extra features. That might be worthwhile if this was code in a package, but it seems strange for a pedagogical example to spend so much time on e.g. argument handling.

dhairyagandhi96:
One model (a new basic model) that shows off these things is fine. Doing it to everything makes it difficult to approach from a new comers standpoint.

MikeInnes:
Yeah – model loading/saving might make sense as a tutorial for example. But all the keyword argument handling makes very little sense to me – these things are meant to be self contained.

PR #228, opened 2020-04-27
Update contrib/diffeq

Emmanuel-R8:


contrib/diffeq/lotka-volterra.jl
```diff
@@ -1,49 +1,126 @@
...
+
+
+# The parameter `loss_data` in `train!` (below) contains a list of data that are provided
+# to the loss function. `train!`` works by iterating over that list. In our case, the
+# calculation of the loss function does not require any additional data.
+# Therefore `loss_data` is filled with a whole bunch of nothing.
+loss_data = Iterators.repeated((), 1000)
+
+# ADAM is a great default optimiser
+optimiser = Flux.ADAM(0.1)
```
CarloLucibello, line 123: Adam standard learning rate is much smaller, 3e-4. Is training with such a high value stable?

CarloLucibello:
this generally looks very good

contrib/diffeq/lotka-volterra.jl
Emmanuel-R8: I kept the value that was there before. This aside, the solution converges after 600/700 iterations. So I left the 1,000 and didn't explore other learning rates.

PR #230, opened 2020-05-08
Adding a seed for sampling

mcognetta:
Adding an optional seed to give context to the text generation step.


dhairyagandhi96:
Can you add that as a kwarg instead

mcognetta:
Done.

PR #249, opened 2020-08-11
Zygote

SubhadityaMukherjee:
Hello! 

I :heart: Flux but the model zoo is a "bit" unstructured. Here is the first step towards making it a bit better. **No code changes.** Just adding notebooks as well as separate jl files and changing the folders to make it easier to go through.

This is the first step so I just added a folder for VGG. The others are in progress. 
I aim to make a **unified way of using Flux** unlike the current main branch where every file has some different implementation altogether. It would be so much easier for **beginners to start.**


DhairyaLGandhi:
Note that the notebooks can be generated automatically from the scripts already present in the repo.

SubhadityaMukherjee:
@DhairyaLGandhi okay. I can remove those if needed then. 
Should I proceed with the folder structure though? I feel ordering them based on models instead of datasets is a better way to go...

DhairyaLGandhi:
I merely meant that it can speed up generating the notebooks and save time and effort. Although some updates would be inevitable.

The structure looks good so far.

CarloLucibello:
Maybe we can add some pluto notebooks, nicer versioning, and close this PR since has been stale for a while

Issue #252, opened 2020-08-19
CuArrays Dependency for 60 Minute Blitz


adinhobl:
In [60-minute-blitz.jl](https://github.com/FluxML/model-zoo/blob/master/tutorials/60-minute-blitz.jl), I'm having difficulty getting both CuArrays and Flux to both run their associated lines and it has to do with dependencies. 

```
using CuArrays
x = cu(rand(5, 3))

f(x) = 3x^2 + 2x + 1

f(5)

using Flux: gradient

df(x) = gradient(f, x)[1]

df(5)

ddf(x) = gradient(df, x)[1]

ddf(5)

```

1. add CuArrays then Flux
    [3a865a2d] CuArrays v2.2.2
    [587475ba] Flux v0.8.3
    If I use these I get `ERROR: Use `gradient(...; nest = true)` for nested derivatives` for the `ddf(5)` line

1. add CuArrays then Flux, then `]up Flux`
    [3a865a2d] CuArrays v2.2.2
    [587475ba] Flux v0.10.4
    I get the same error.

1. add Flux only
    [587475ba] Flux v0.11.1
    I don't get that nested gradient error but any attempt to add CuArrays downgrades Flux

Everything in the file seems to work if I replace `using CuArrays` with `using CUDA` and the fresh default packages work out by default so that the gradients no longer say `(tracked)`. I'm not sure if there is any performance or behind-the-scenes difference though. 
  [052768ef] CUDA v1.2.1
  [587475ba] Flux v0.11.1

You could also change the section at lines 276 and 277 to remove `using Zygote` and `using CuArrays`. 

Related discussion #https://github.com/SciML/DiffEqFlux.jl/issues/390

DhairyaLGandhi:
Yeah the environment needs updating, a PR doing that would be well appreciated

adinhobl:
I'll submit one soon with the changes I mentioned above.

PR #253, opened 2020-08-20
Updated to CUDA from CuArrays dependency

adinhobl:
Removed dependency on CuArrays as it was causing conflicts with Flux and erroring. Described more in #252.  

Updated description to refer to CUDA.

Eliminated `using Zygote` as it's not needed explicitly at any point for Flux to train the model.

### DIFF
--- a/tutorials/60-minute-blitz.jl
+++ b/tutorials/60-minute-blitz.jl
@@ -96,14 +96,14 @@ W * x
 
 # ### CUDA Arrays
 
-# CUDA functionality is provided separately by the [CuArrays
-# package](https://github.com/JuliaGPU/CuArrays.jl). If you have a GPU and CUDA
-# available, you can run `] add CuArrays` in a REPL or IJulia to get it.
+# CUDA functionality is provided separately by the [CUDA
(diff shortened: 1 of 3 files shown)


adinhobl:
@nirmal-suthar, I updated the Project.toml file to include `Statistics` and `Images`. I also ran `] up` to update the Manifest.toml. Let me know if anything else is needed.

DhairyaLGandhi:
If we are removing CUDA from Project.toml, could we add in a note to the commented `using CUDA` line to include a note saying you should add it in. `using Pkg; Pkg.add("CUDA")` or similar

adinhobl:
@DhairyaLGandhi Neither CUDA nor CuArrays were originally part of the project. I added CUDA in while I was testing it out, so that commit is bringing it back in line closer to how it was originally.

Lines 99-101 of the 60 minute blitz were previously updated to say this:
```
# CUDA functionality is provided separately by the [CUDA
# package](https://github.com/JuliaGPU/CUDA.jl). If you have a GPU and CUDA
# available, you can run `] add CUDA` in a REPL or IJulia to get it.
```
Do we feel like that is sufficient, or would we rather just include it?

DhairyaLGandhi:
This should be fine, thanks all!

PR #257, opened 2020-09-28
Changed dependency from CuArrays to CUDA

pritamd47:
[CuArrays.jl is deprecated](https://github.com/JuliaGPU/CuArrays.jl#cuarraysjl-deprecated-use-cudajl-instead) and is replaced by CUDA.jl. Changing `CuArrays` to `CUDA` fixes errors related to this.

### DIFF
--- a/vision/cdcgan_mnist/cGAN_mnist.jl
+++ b/vision/cdcgan_mnist/cGAN_mnist.jl
@@ -12,8 +12,8 @@ using CUDAapi
 using Zygote
 if has_cuda()		# Check if CUDA is available
     @info "CUDA is on"
-    import CuArrays		# If CUDA is available, import CuArrays
-    CuArrays.allowscalar(false)
+    import CUDA		# If CUDA is available, import CUDA
+    CUDA.allowscalar(false)
 end
 
 @with_kw struct HyperParams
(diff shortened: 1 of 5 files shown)


DhairyaLGandhi:
Worth doing a sweep of the other models in the directory, if you're up for it

pritamd47:
Absolutely, I've updated rest of the files still using CuArrays. 
But I notice many `Project.toml` files still have CuArrays as dependencies. If I am not wrong, these too need updating right? Let me know if that's the case, and I'll change those as well.

ToucheSir:
Apologies if I'm hijacking this thread, but do these models still function after upgrading to CUDA? Presumably everything would have to be updated to Flux 0.11 simultaneously? This seems to have stalled many similar PRs already, so ideally this one doesn't suffer the same fate. :)

pritamd47:
I think that's a very valid point. I am new to Flux and used model-zoo for getting started. I have to admit, I do not know enough about either model-zoo or its dependencies to comment about the issue, but what worked for me was just removing CuArrays from the environment, and adding the latest CUDA.jl (I have only tested the MLP.jl script). 

I hope someone more experienced than me can comment on this issue.

joostveenema:
All models have been moved from CuArrays to CUDA on the current master branch

PR #258, opened 2020-10-16
Updated Dependencies for MNIST Examples

adinhobl:
I updated the dependencies for other MNIST Vision examples. Everything trained and printed out nicely.

For many of the files that download datasets from `MLDatasets`, I couldn't get the example to run top to bottom without having to intervene manually and go download the dataset myself. The `getdata` function in mlp.jl, for example, only works if you already have the dataset, otherwise it prints out terms and conditions in the prompt, but it's too late to respond.

I got around this by adding something to manually download the dataset before calling `train()` but I realize this isn't a very clean way to do it. Open to suggestions.


adinhobl:
bump :)

DhairyaLGandhi:
I think this is good, thanks!

PR #262, opened 2020-11-16
Avoid repeating download

Eggiverse:
In example of [MNIST mlp](https://github.com/FluxML/model-zoo/blob/0e8de61cf466868b13a629e9a9275cfe7eafa6c6/vision/mnist/mlp.jl#L21),
original data is always downloaded even if it is already in the local machine.

This pull request fixes this issue while bypassing the confirmation under the guidance [here](https://www.oxinabox.net/DataDeps.jl/stable/z10-for-end-users/#Configuration-1).

### DIFF
--- a/vision/mnist/mlp.jl
+++ b/vision/mnist/mlp.jl
@@ -18,7 +18,7 @@ end
 end
 
 function getdata(args)
-    MLDatasets.MNIST.download(i_accept_the_terms_of_use=true)
+    ENV["DATADEPS_ALWAYS_ACCEPT"] = "true"
 
     # Loading Dataset	
     xtrain, ytrain = MLDatasets.MNIST.traindata(Float32)


DhairyaLGandhi:
How does this avoid downloading the dataset multiple times? Maybe we should check for the dataset and download if it isn't present?

Eggiverse:
`MLDataset.MNIST.download` eventually calls `download` defined in `DataDeps` which [aims for debugging](https://www.oxinabox.net/DataDeps.jl/stable/z20-for-pkg-devs/#download-for-low-level-programmatic-resolution.-1). `DataDeps` works in a way that whenever you acquire any data, it automatically checks if it is present and download it if not. As a result, explicitly calling `download` is unnecessary. 

The only reason I can think of calling `download` here is to avoid downloading confirmation, and for this, `DataDeps` suggests [setting environment variable `DATADEPS_ALWAYS_ACCEPT` to be `"true"`](https://www.oxinabox.net/DataDeps.jl/stable/z10-for-end-users/#Configuration-1).

CarloLucibello:
> How does this avoid downloading the dataset multiple times? Maybe we should check for the dataset and download if it isn't present?

The download will be triggered by `MNIST.traindata` only if needed

PR #263, opened 2020-11-29
Updated cGan_mnist example dependencies, some issues still

adinhobl:
I updated the dependencies for the cdcgan_mnist example to be more modern.

Three other things:

1. I still occasionally get an warning like below. It doesn't happen every time I run it, but sometimes it does. I don't know if there is something for me to fix.
```
[ Info: Epoch 1
┌ Warning: calls to Base intrinsics might be GPU incompatible
│   exception =
│    You called log1p(x::Float32) in Base.Math at special/log.jl:360, maybe you intended to call log1p(x::Float32) in CUDA at /home/andrew/.julia/packages/CUDA/YeS8q/src/device/intrinsics/math.jl:85 instead?
│    Stacktrace:
│     [1] log1p at special/log.jl:360
│     [2] broadcast_kernel at /home/andrew/.julia/packages/GPUArrays/jhRU7/src/host/broadcast.jl:59
└ @ GPUCompiler ~/.julia/packages/GPUCompiler/uTpNx/src/irgen.jl:68
[ Info: Train step 0, Discriminator loss = 1.3628006, Generator loss = 0.6890712
[ Info: Epoch 2
[ Info: Epoch 3
[ Info: Train step 1000, Discriminator loss = 1.1395278, Generator loss = 1.0239918
```
2. I included a line in the `load_data` function to actually download the data. Without it, the prompt to download the data will appear, but I don't think there's a way for me to respond, so it gets stuck. The way I have it now, it will download data every time the script is run, which isn't optimal either.
```
function load_data(hparams)
    MLDatasets.MNIST.download(i_accept_the_terms_of_use=true)
```
3. I also removed the model output images from the repository. I'm not sure if it's better to have them or not, open to suggestions.


CarloLucibello:
Fix for 1.: In newer versions of julia, loss functions should not be broadcasted and also perform reduction (mean by default). So                                `mean(logitbinarycrossentropy.(fake_output, 0f0))` becomes `logitbinarycrossentropy(fake_output, 0f0)`

vision/cdcgan_mnist/cGAN_mnist.jl
```diff
@@ -75,6 +76,8 @@ function (m::Generator)(x, y)
 end
 
 function load_data(hparams)
+    MLDatasets.MNIST.download(i_accept_the_terms_of_use=true)
```
CarloLucibello, line 79: let's remove this. It's not needed when including the script from a REPL (julia's typical usage)

CarloLucibello:
> 3. I also removed the model output images from the repository. I'm not sure if it's better to have them or not, open to suggestions.

Let's keep them! Or even better, replace them with the ones you obtain from a run on your machine. Keeping the generated images for comparison safeguards us from visual rquality regressions

vision/cdcgan_mnist/cGAN_mnist.jl
adinhobl: Will do. As someone who ran this using the VSCode extension, it just printed out the text and there was no way for me to respond that "yes please download it". I suspect that anyone who doesn't use the REPL by default (mostly newer users) could stumble in the same spot. Should I add a comment?

adinhobl: That probably also explains what `cd(@__DIR__)` is doing in all these files :sweat_smile: excuse my ignorance

adinhobl:
Note that in the most recent commits I had to change the global `.gitignore`. Somehow the previous output images were included in the repo, although both the global `.gitignore` and the one in the cdcgan folder both excluded the `output` folder... Not sure how they ended up there to begin with. The above commit spam was me figuring that out.

adinhobl:
bump :)

CarloLucibello:
thank you for this!

Issue #266, opened 2021-01-17
Refresh model zoo


joost-ve:
I started testing all the models in this repo against the flux 0.11.4 release on julia 1.5 (ref https://github.com/FluxML/ML-Coordination-Tracker/issues/9). In this issue i will collect all encoutered problems and try to list all other open issues for each model. The tests are ran against my [fork](https://github.com/joost-ve/model-zoo) which contains the updated manifests and hotfixes needed to actually run some of the models.

**all**
- [ ] Parameters.@with_kw vs Base.@kwdef, see https://github.com/mauro3/Parameters.jl/issues/89. None of the models seem to depend on features not provided by the Base implementation.
- [ ] BSON.jl seems not really well maintained and several models will not be able to use it on julia 1.6 unless this [issue](https://github.com/JuliaIO/BSON.jl/issues/80) is fixed.

**vision/cdcgan_mnist**:
- runs without issues

**vision/cifar10**:
- [ ] dependency problems when updateing to Flux 0.11.4
- [ ] depends on Metalhead dataset instead of MLDatasets

**vison/cppn**:
- [ ] no GPU support

**vison/dcgan_mnist**:
- [ ] fails, probably this [issue](https://github.com/FluxML/Flux.jl/issues/1350)
- [ ] hparams.latent_dim not available in generator constructor [(hotfix)](https://github.com/joost-ve/model-zoo/commit/eb7fd41221bd6650cc5d06d5352d79a68f9e0e6f)
- [ ] uses deprecated logitbinarycrossentropy [(hotfix)](https://github.com/joost-ve/model-zoo/commit/515a81fd5eb988850da5511c5526cca082271e0d)
- [ ] fails on julia 1.6-beta1 because of BSON.jl issue

**vison/lenet_mnist**:
- [ ] depends on CUDAapi instead of CUDA
- [ ] fails on julia 1.6-beta1 because of BSON.jl issue [(hotfix)](https://github.com/joost-ve/model-zoo/commit/ff733cea0f850c487156b89ff2910000a6edc86f)

**vison/mnist/mlp**:
- runs without issues

**vison/mnist/conv**:
- [ ] depends on Flux.Data.MNIST instead of MLDatasets

**vison/mnist/autoencoder**:
- [ ] depends on Flux.Data.MNIST instead of MLDatasets

**vison/vae_mnist**:
[...]

PR #267, opened 2021-01-17
Fix vae_mnist model

joost-ve:
- Replaces CUDAapi with CUDA
- Replaces Flux.logitbinarycrossentropy with Flux.Losses.logitbinarycrossentropy
- Makes the Encoder a functor so will actually be moved to the CPU before saving the BSON file

see also https://github.com/FluxML/model-zoo/issues/266


vision/vae_mnist/vae_mnist.jl
```diff
@@ -6,10 +6,11 @@
 
 using Base.Iterators: partition
 using BSON
-using CUDAapi: has_cuda_gpu
+using CUDA
 using DrWatson: struct2dict
 using Flux
-using Flux: logitbinarycrossentropy, chunk
+using Flux: @functor, chunk
```
CarloLucibello, line 12: looks like chunk is not used

vision/vae_mnist/vae_mnist.jl
```diff
@@ -6,10 +6,11 @@
 
 using Base.Iterators: partition
```
CarloLucibello, line 7: partition not used

CarloLucibello:
Is generation fine or producing grey images?

vision/vae_mnist/vae_mnist.jl
joost-ve: It is used in the convert_to_image function: https://github.com/FluxML/model-zoo/pull/267/files#diff-d2fcc7489d82937b5b466b70d629d200f23003b00a0e923cfa5031eedfa54d4cL70

joost-ve:
Just grey images

joost-ve:
Runs without issues and resulting images are ok now. It could use some improvement overall, but I think we should just leave it like this for now.

vision/vae_mnist/vae_mnist.jl
```diff
@@ -59,15 +61,15 @@ function model_loss(encoder, decoder, λ, x, device)
     # KL-divergence
     kl_q_p = 0.5f0 * sum(@. (exp(2f0 * logσ) + μ^2 -1f0 - 2f0 * logσ)) / len
 
-    logp_x_z = -sum(logitbinarycrossentropy.(decoder_z, x)) / len
+    logp_x_z = -logitbinarycrossentropy(decoder_z, x, agg=sum) / len
```
CarloLucibello, line 64: So this was causing the grey images? I would have thought that 
```julia
sum(logitbinarycrossentropy.(decoder_z, x))
```
and
```julia
logitbinarycrossentropy(decoder_z, x, agg=sum)
```
produced the same output...

CarloLucibello: ah now I see the problem was in a misplaced sigmoid

PR #268, opened 2021-01-17
Fix lenet_mnist model

joost-ve:
- Replace CUDAapi with CUDA
- Workaround to fix saving the model on julia 1.6-beta1 (https://github.com/JuliaIO/BSON.jl/issues/80)

Some small less urgent changes which I now see I added to the commits:
- Removes some unused dependencies from Project.toml
- Parameters.@with_kw replaced with Base.@kwdef

see https://github.com/FluxML/model-zoo/issues/266

### DIFF
--- a/vision/lenet_mnist/lenet_mnist.jl
+++ b/vision/lenet_mnist/lenet_mnist.jl
@@ -23,12 +23,11 @@ function LeNet5(; imgsize=(28,28,1), nclasses=10)
     out_conv_size = (imgsize[1]÷4 - 3, imgsize[2]÷4 - 3, 16)
     
     return Chain(
-            x -> reshape(x, imgsize..., :),
             Conv((5, 5), imgsize[end]=>6, relu),
             MaxPool((2, 2)),
             Conv((5, 5), 6=>16, relu),
             MaxPool((2, 2)),
-            x -> reshape(x, :, size(x, 4)),
+            flatten,
             Dense(prod(out_conv_size), 120, relu), 
             Dense(120, 84, relu), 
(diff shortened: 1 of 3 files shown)


CarloLucibello:
Looks good, thanks! Maybe we could add proper compat bounds to each model zoo example, so that the last setup on which it was tested is more clear, without having to dig into the manifest

vision/lenet_mnist/lenet_mnist.jl
```diff
@@ -23,12 +22,11 @@ function LeNet5(; imgsize=(28,28,1), nclasses=10)
     out_conv_size = (imgsize[1]÷4 - 3, imgsize[2]÷4 - 3, 16)
```
darsnack, line 22: Not sure if the intent is just bug fixes or a refresh. If it is the former, then ignore this comment. If it is the latter, then we should use `Flux.outputsize` here instead of manually calculating the output size.

joost-ve: This pull request was just to get the model running on the next julia release. I will add your comment to the list in the tracking issue though.

joost-ve:
Compat bounds sounds like a great idea! How aggressive should we be here? Is it ok to add Julia 1.5 and Flux 0.11.4 as minumum versions (these are the versions which i use for testing)?

CarloLucibello:
Since we don't have CI here, I would be super aggressive and use the versions in your testing configuration.

joost-ve:
I added compat entries for julia, flux and cuda. Runs without issues now.

PR #269, opened 2021-01-17
Fix dcgan_mnist model

joost-ve:
Fixes:
- Replace Flux.logitbinarycrossentropy with Flux.Losses.logitbinarycrossentropy
- Take the latent_dim as input to the generator constructor

No yet fixed:
- Crashes, likely because of this issue : https://github.com/FluxML/Flux.jl/issues/1350

see https://github.com/FluxML/model-zoo/issues/266

### DIFF
--- a/vision/dcgan_mnist/dcgan_mnist.jl
+++ b/vision/dcgan_mnist/dcgan_mnist.jl
@@ -56,12 +56,12 @@ end
 
 # Loss functions
 function discriminator_loss(real_output, fake_output)
-    real_loss = mean(logitbinarycrossentropy(real_output, 1f0))
-    fake_loss = mean(logitbinarycrossentropy(fake_output, 0f0))
+    real_loss = logitbinarycrossentropy(real_output, 1)
+    fake_loss = logitbinarycrossentropy(fake_output, 0)
     return real_loss + fake_loss
 end
 
-generator_loss(fake_output) = mean(logitbinarycrossentropy(fake_output, 1f0))
+generator_loss(fake_output) = logitbinarycrossentropy(fake_output, 1)
 
 function train_discriminator!(gen, dscr, x, opt_dscr, hparams)
     noise = randn!(similar(x, (hparams.latent_dim, hparams.batch_size))) 
--- a/vision/dcgan_mnist/dcgan_mnist.jl
+++ b/vision/dcgan_mnist/dcgan_mnist.jl
@@ -8,6 +8,7 @@ using Statistics
 using Parameters: @with_kw
 using Printf
 using Random
+using CUDA
 
 @with_kw struct HyperParams
     batch_size::Int = 128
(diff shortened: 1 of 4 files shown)


vision/dcgan_mnist/dcgan_mnist.jl
```diff
@@ -56,12 +56,12 @@ end
 
 # Loss functions
 function discriminator_loss(real_output, fake_output)
-    real_loss = mean(logitbinarycrossentropy.(real_output, 1f0))
-    fake_loss = mean(logitbinarycrossentropy.(fake_output, 0f0))
+    real_loss = mean(logitbinarycrossentropy(real_output, 1f0))
+    fake_loss = mean(logitbinarycrossentropy(fake_output, 0f0))
```
darsnack, line 60: As a best practice, we should use `one(eltype(real_output))` and `zero(eltype(real_output))`, etc. Not a big deal for this function, but new users often get bit by `Float32`/`Float64` mismatch elsewhere, so we should make our sample code illustrate how to always match the argument types.

vision/dcgan_mnist/dcgan_mnist.jl
```diff
@@ -56,12 +56,12 @@ end
...
     return real_loss + fake_loss
 end
 
-generator_loss(fake_output) = mean(logitbinarycrossentropy.(fake_output, 1f0))
+generator_loss(fake_output) = mean(logitbinarycrossentropy(fake_output, 1f0))
```
darsnack, line 64: Same here.

vision/dcgan_mnist/dcgan_mnist.jl
CarloLucibello:
```suggestion
    real_loss = logitbinarycrossentropy(real_output, 1)
    fake_loss = logitbinarycrossentropy(fake_output, 0)
```
All losses now perform `mean` reduction by default. Integers are also fine in this case, otherwise @darsnack could be shortened a bit with `one(real_output)`

joost-ve: Fixed these, but the suggested `one(real_output)` does not work for me (`one` is only implemented for scalars or square matrices).

joost-ve:
I removed the tanh activation of the last `ConvTranspose` layer and added it as an seperate layer, this works around the issue that triggered the crash. Scripts runs without problems now, output after 20 epochs seems ok.

vision/dcgan_mnist/dcgan_mnist.jl
```diff
@@ -105,7 +106,7 @@ function train(; kws...)
     dscr = Discriminator() |> gpu
```
CarloLucibello, line 106: define a `device` instead of just using `gpu`?

joost-ve: Makes sense

CarloLucibello:
I think another manifest got caught in the update, it doesn't really matter although it will conflict with the other PR. 
I'll merge whenever you say this is ready in any case

joost-ve:
Ah yes, I missed the extra 'c' in the filename. The cdcgan models runs fine with the updated manifest. 

The biggest problems with this model are solved I think, so LGTM

PR #270, opened 2021-01-17
Fix cifar10 model

joost-ve:
- Replaces dependencies which prevented updating to flux 0.11.4
- Wrote custom training loop to allow data to be moved to the GPU in batches, this allows running the model on lower end cards without OOM errors.

### DIFF
--- a/vision/cifar10/cifar10.jl
+++ b/vision/cifar10/cifar10.jl
@@ -5,11 +5,10 @@ using Parameters: @with_kw
 using Images: channelview
 using Statistics: mean
 using Base.Iterators: partition
-using CUDAapi
+using CUDA
 if has_cuda()
     @info "CUDA is on"
-    import CuArrays
-    CuArrays.allowscalar(false)
+    CUDA.allowscalar(false)
 end
 
 @with_kw mutable struct Args
--- a/vision/cifar10/cifar10.jl
+++ b/vision/cifar10/cifar10.jl
@@ -44,94 +44,104 @@ end
 
 # VGG16 and VGG19 models
 function vgg16()
-    return Chain(
-            Conv((3, 3), 3 => 64, relu, pad=(1, 1), stride=(1, 1)),
-            BatchNorm(64),
-            Conv((3, 3), 64 => 64, relu, pad=(1, 1), stride=(1, 1)),
-            BatchNorm(64),
-            MaxPool((2,2)),
-            Conv((3, 3), 64 => 128, relu, pad=(1, 1), stride=(1, 1)),
-            BatchNorm(128),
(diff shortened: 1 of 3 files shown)


vision/cifar10/cifar10.jl
```diff
@@ -158,21 +145,49 @@ function train(; kws...)
...
+
+            gs = Flux.gradient(ps) do 
+                loss(x, y)
+            end
+
+            Flux.update!(opt, ps, gs)
+        end
+
+        validation_loss = 0f0
+        for (x, y) in val_data
```
DhairyaLGandhi, line 166: Validation should also be batched I think

joost-ve: What do you mean? val_data is a batched data loader, same as train_data, see line 139.

vision/cifar10/cifar10.jl
```diff
@@ -20,40 +20,26 @@ end
...
+    val_x = float(val_x)
+    val_y = onehotbatch(val_y, 0:9)
+    
+    return (train_x, train_y), (val_x, val_y)
+end
 
-    test = (testX,testY)
-    return test
+function get_test_data()
+    test_x, test_y = CIFAR10.traindata()
```
CarloLucibello, line 37:
```suggestion
    test_x, test_y = CIFAR10.testdata()
```

vision/cifar10/cifar10.jl
```diff
@@ -158,21 +145,49 @@ function train(; kws...)
...
+    # Defining the optimizer
     opt = ADAM(args.lr)
+    ps = Flux.params(m)
+
     @info("Training....")
     # Starting to train models
-    Flux.@epochs args.epochs Flux.train!(loss, params(m), train, opt, cb = evalcb)
+    for epoch in 1:args.epochs
+        for (x, y) in train_data
+            x, y = gpu(x), gpu(y)
```
CarloLucibello, line 156: define a `device` instead of using `gpu`?

vision/cifar10/cifar10.jl
joost-ve: Oops :wink:

vision/cifar10/cifar10.jl
```diff
@@ -20,159 +20,185 @@ end
...
+
+            Flux.update!(opt, ps, gs)
+        end
+
+        validation_loss = 0f0
+        for (x, y) in val_data
+            x, y = x |> device, y |> device
+            validation_loss += loss(x, y)
+        end
+        @show validation_loss
```
CarloLucibello, line 180:
```suggestion
         validation_loss /= length(val_data)
        @show validation_loss
```

PR #271, opened 2021-01-19
Fix treebank model

joostveenema:
Several errors in the sscript prevented it from running. Script runs fine now and loss seems to converge, but not 100% sure if the model is now how it was supposed to be..


DhairyaLGandhi:
Lgtm, thanks!

PR #273, opened 2021-01-20
Fix housing model

joostveenema:
This model had a diverging loss due to a minus sign in the update step. This PR does also replace the custom meansquareerror function with `Flux.Losses.mse`.

### DIFF
--- a/other/housing/housing.jl
+++ b/other/housing/housing.jl
@@ -53,8 +53,10 @@ end
 # Function to predict output from given parameters
 predict(x, m) = m.W*x .+ m.b
 
-# Mean Squared Error
-meansquarederror(ŷ, y) = sum((ŷ .- y).^2)/size(y, 2)
+# Define the mean squared error function to be used in the loss 
+# function. An implementation is also available in the Flux package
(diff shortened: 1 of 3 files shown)


other/housing/housing.jl
```diff
@@ -67,24 +65,24 @@ function train(; kws...)
...
 
     for i = 1:500
       g = gradient(() -> loss(x_train, y_train), θ)
       for x in θ
-        update!(x, -g[x]*η)
+            update!(x, g[x]*η)
       end
       if i%100==0
           @show loss(x_train, y_train)
         end
```
CarloLucibello, line 81: indentation issues

CarloLucibello:
This example seems to be quite low level, maybe we should leave the custom mse there?

joostveenema:
It is. But I think it makes more sense to show using a custom loss function in of the tutorials were it can not simply be replaced by one from the available functions in Flux. Removed it from this PR, but left the in the list of issues for now.

DhairyaLGandhi:
If we add some comments that bring attention to the custom mse, that makes the same point, but tutorials with more complex loss functions are welcome of course

PR #275, opened 2021-01-21
Fix transfer learning tutorial

joostveenema:
- Use correct path for loading dataset
- Adds missing packages to the project environment
- Removes some unused imports


DhairyaLGandhi:
Thanks!

PR #276, opened 2021-01-23
add dataloader example

lilianabs:


DhairyaLGandhi:
Hey @lilianabs could you follow up with moving it to `tutorials` instead of `other`? Thanks

DhairyaLGandhi:
Thanks for this!

PR #277, opened 2021-01-23
Add compats

joostveenema:
Adds a [compat] section to the Project.toml files for the models. The models are tested to work without errors on these versions.


DhairyaLGandhi:
Brilliant, although not sure if Pkg necessarily respects that for non package Project.tomls, but good to have.

PR #278, opened 2021-01-23
Add Fluxbot

DhairyaLGandhi:
With this, we should be able to listen for certain commands on issues on the repos we care for. For example, if we wish to test a Flux PR/ branch/ commit to see how it effects models in the model-zoo. We can add some CI to the zoo, and on similar grounds convert the models into markdown/ HTML via Literate.jl to publish on the website automatically. The bot would respond with status of the jobs when they finish, and even put the output in a gist. It would respond to `@ModelZookeeper`, but only once we merge this so GA can pick up the workflow (ah well). We would need a similar `.fluxbot` in the repos we want it to listen to, or can make it into a Github app, this seemed easier for now.

Of course, we can optionally give it certain models or directories (eg. `build MNIST`) so its not too long. It would use GPUs wherever possible, but that's dependent on how the model is written.

We can extrapolate to also bring Metalhead under the purview so we can test with more standard models as well.

There is some config remaining - basically having a pipeline on Buildkite rather than on GitLab and adding the relevant keys and tokens.

TODO:

- [ ] Move this to Buildkite


DhairyaLGandhi:
@ModelZookeeper commands

CarloLucibello:
Can you add a brief description in the OP of what this does?

DhairyaLGandhi:
Done, lmk if it needs more description.

PR #279, opened 2021-01-23
Missing commit to fix cifar10 example

joostveenema:
I apparently forget to commit these lines. If not included the script complains about the missing device function This is not the nicest solution but it does at least fix the script.

### DIFF
--- a/vision/cifar10/cifar10.jl
+++ b/vision/cifar10/cifar10.jl
@@ -134,14 +134,6 @@ function train(; kws...)
     # Initialize the hyperparameters
     args = Args(; kws...)
 	
-    if CUDA.has_cuda()
-        device = gpu
-        @info "Training on GPU"
-    else
-        device = cpu
-        @info "Training on CPU"
-    end
-    
     # Load the train, validation data 
     train, val = get_processed_data(args)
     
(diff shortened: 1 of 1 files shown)


CarloLucibello:
what do you mean it is not the nicest solution? you have in mind a different code pattern to send models to cpu/gpu?

joostveenema:
It now has the following lines in both the train and test function:

```
    if CUDA.has_cuda()
         device = gpu
     else
         device = cpu
     end
```

I think there should only be one check for a GPU. But this is probably something to keep in mind when rewriting the model in a tutorial/notebook style and just leave it like this for now.

CarloLucibello:
`device = CUDA.has_cuda() ? gpu : cpu` in the global space and pass it to test and train?

DhairyaLGandhi:
I agree it's ugly and adds nothing. `gpu` is a no-op when there isn't a working GPU available, so you don't have to have it here at all.

joostveenema:
Just using `gpu` and removing the device handling seems like to the way to go here. Updated the PR.

CarloLucibello:
i find it better to be explicit, but it's not important

PR #283, opened 2021-01-27
Update Metalhead

joostveenema:
Update metalhead dep for 60 minute blitz and transfer learning tutorials. In the 60 minute blitz it does only use it for the cifar10 dataset, rewriting the tutorial to use MLDatasets will be done later.


DhairyaLGandhi:
Thanks!

PR #288, opened 2021-02-10
Add ImageIO dependency required for autoencoder example

c42f:
Without this, the autoencoder example fails on the last line.


DhairyaLGandhi:
Thanks!

PR #292, opened 2021-02-20
Flux v0.12 and reorg

CarloLucibello:


DhairyaLGandhi:
Why are you doing this? Esp the dP examples are now gone?

CarloLucibello:
some rationalizing, too much rationalizing indeed, I'll close this

PR #299, opened 2021-04-05
Update conv_mnist.jl

batate:
correct grammar


DhairyaLGandhi:
Seem great! Thanks! I quickly cleared out the differences in fonts.

PR #315, opened 2021-09-14
switch dot/map to list comprehension for RNN eval

albheim:
Fixed a small issue that came up [here](https://discourse.julialang.org/t/issues-with-recurrent-layers-in-flux-jl/67815/9) on discourse about using dot notation over RNN models. Switched to use list comprehension to make sure the evaluation order is correct.


CarloLucibello:
thanks!

PR #319, opened 2021-10-10
Avoid mixing of types in DCGAN

### LINKED ISSUE #318: Slow fallback implementation invoked in DCGAN MNIST example
The [dcgan_mnist tutorial](https://github.com/FluxML/model-zoo/blob/master/vision/dcgan_mnist/dcgan_mnist.jl) throws warning:

```
┌ Warning: Slow fallback implementation invoked for ∇conv_data!  You probably don't want this; check your datatypes.
│   yT = Float32
│   T1 = Float32
│   T2 = Float64
└ @ NNlib ~/.julia/packages/NNlib/P9BhZ/src/conv.jl:291
``` 
This is beacuse the weights for the ConvTranspose layers are accidentally set to Float64 due to the weight init function. This can be simply fixed by using 0.02f0(Float32) instead of 0.02(Float64). To be precise, the [dcgan_init](https://github.com/FluxML/model-zoo/blob/d1a598a5f5b634f765aa58c4d963925e77f05264/vision/dcgan_mnist/dcgan_mnist.jl#L35) function should be,
```julia
dcgan_init(shape...) = randn(Float32, shape...) * 0.02f0
```

Dsantra92:
Closes #318
I found two places where F64 are creating issues.

### DIFF
--- a/vision/dcgan_mnist/dcgan_mnist.jl
+++ b/vision/dcgan_mnist/dcgan_mnist.jl
@@ -32,7 +32,7 @@ end
 
   
 # weight initialization as given in the paper https://arxiv.org/abs/1511.06434
-dcgan_init(shape...) = randn(Float32, shape...) * 0.02
+dcgan_init(shape...) = randn(Float32, shape...) * 0.02f0
 
 function Discriminator()
     return Chain(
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Definitely an improvement!

PR #324, opened 2021-11-22
Update dataloader.jl with comments

logankilpatrick:
I am working on a Transfer Learning Tutorials and was using this code so I decided to port over my comments which will hopefully help make this more accessible

### DIFF
--- a/tutorials/transfer_learning/dataloader.jl
+++ b/tutorials/transfer_learning/dataloader.jl
@@ -26,7 +26,7 @@ function load_batch(n = 10, nsize = (224,224); path = PATH)
   labels = map(x -> occursin("dog.",x) ? 1 : 0, imgs_paths)
   # Here, dog is set to 1 and cat to 0
   
-  # Convert the text based names to 0 or 1
+  # Convert the text based names to 0 or 1 (one hot encoding)
   labels = Flux.onehotbatch(labels, [0,1])
   
   # Load all of the images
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
It's supposed to output a batch of  images and their labels. The first element is the data and the second is the one hot style labels. Half the images are always dogs and the other half are cats, that was just so the demo trained properly because we don't have too much data and I didn't want to risk over fitting on a class. We can do some class balance techniques to ensure we are sampling all classes fairly, which I think would be a good addition.

tutorials/transfer_learning/dataloader.jl
```diff
@@ -7,16 +7,41 @@ if isempty(readdir(PATH))
...
+# Get all of the files with "dog" in the name
 const DOGS = filter(x -> occursin("dog", x), FILES)
+
+# Get all of the files with "cat" in the name
 const CATS = filter(x -> occursin("cat", x), FILES)
 
+# Takes in the number of requested images per batch ("n") and image size
+# Returns a 4D array with images and an array of labels
 function load_batch(n = 10, nsize = (224,224); path = PATH)
+  if ((batchsize % 2) != 0)
```
DhairyaLGandhi, line 19: Is that really the case?

DhairyaLGandhi:
Lgtm for the most part! Thank you!

tutorials/transfer_learning/dataloader.jl
logankilpatrick: Yes, it errors if you pass in 1 or 3 for example

DhairyaLGandhi: Maybe we need to use `\div` instead of `/` in line 23 (`img_paths = ...`)

logankilpatrick: The same error occurs even when using that division sign

DhairyaLGandhi:
May I ask what the error was?

logankilpatrick:
This is what happens when I pass in `load_batch(1)`

tutorials/transfer_learning/dataloader.jl
```diff
@@ -7,16 +7,46 @@ if isempty(readdir(PATH))
...
+      print("Batch size must be an even number")
+  end
+  # Sample N dog images and N cat images, shuffle, and then combine them into a batch
   imgs_paths = shuffle(vcat(sample(DOGS, Int(n/2)), sample(CATS, Int(n/2))))
```
darsnack, line 14: If `n == 1`, then this will produce an empty array for `img_paths`. It would be safer to do

```julia
ndogs = div(n, 2)
ncats = n - ndogs
img_paths = shuffle(vcat(sample(DOGS, ndogs), sample(CATS, ncats)))
```

You would also be able to remove the even restriction.

PR #326, opened 2021-11-27
Fix error when running char-rnn.jl

logankilpatrick:
Right now, I get the following when trying to run char-rnn.jl: 

```Julia
ERROR: LoadError: error in method definition: function StatsBase.sample must be explicitly imported to be extended
Stacktrace:
 [1] top-level scope
```

Updating the function name resolves this issue.

### DIFF
--- a/text/char-rnn/char-rnn.jl
+++ b/text/char-rnn/char-rnn.jl
@@ -72,7 +72,7 @@ function train(; kws...)
 end
 
 # Sampling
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):

DhairyaLGandhi (approved):

DhairyaLGandhi:
Thanks! This was helpful!

Issue #327, opened 2021-12-09
Minor error in the 60 minute blitz tutorial


rossinerbe:
Hey, just an FYI that the "Deep Learning with Flux - A 60 Minute Blitz" tutorial has a minor error.

When the test data is being prepped for the convolutional net the code used is this:

```
test_x, test_y = CIFAR10.testdata(Float32)
test_labels = onehotbatch(test_y, 0:9)

test = gpu.([(test_x[:,:,:,i], labels[:,i]) for i in partition(1:10000, 1000)])
```

in the third line the `labels` should be `test_labels` otherwise the labels are from the training rather than the test data, which naturally gives pretty random accuracy results later on.

ToucheSir:
Good catch, mind filing a PR?

rossinerbe:
Ah, actually it looks like the error is just in the html of the tutorial at https://fluxml.ai/tutorials/2020/09/15/deep-learning-flux.html, the .jl file in the repository is fine.

ToucheSir:
Yes, that's because the file in this repo is out of date and hasn't been updated to match https://github.com/FluxML/fluxml.github.io/pull/118. So it'd be a 2-step process: typo fix there, and full update here.

rossinerbe:
#328 Gotcha, submitted the PR.

PR #328, opened 2021-12-10
Update 60-minute-blitz.jl

rossinerbe:
Updated to match the version at https://github.com/FluxML/fluxml.github.io, with the minor error fixed.


DhairyaLGandhi:
Thanks for taking this up! That's a good eye, too. A couple of notes from my end. This is a Julia file (as opposed to the markdown one in the website) so we will definitely need to retain the comments etc. We'd also preferably retain the literature as is. Keeping the changes to fix #327 is a great idea though!

rossinerbe:
Gotcha, changed it to a .jl file with the same changes.

PR #330, opened 2022-01-06
vision : Convmixer architecture

SimonCoste:
I added a minimal example with the ConvMixer architecture trained on CIFAR10. The code is really as minimal as it can be.


DhairyaLGandhi:
Thanks! This is pretty neat! Could we make it so we don't need to rely on a a specific environment rather use the Project and Manifest at the top of the repo? (That could do with an `]up`)

SimonCoste:
Sure. Do I need to make these changes ?

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -0,0 +1,131 @@
...
+end
+
+function ℓ(dataloader, model, device)
+    
+    n = 0
+    loss = 0.0f0
+    acc = 0.0f0
+
+    for (x,y) in dataloader
+        x,y = x |> device, y |> device
```
DhairyaLGandhi, line 61: This doesn't seem efficient, could we move the data prior to the loading? Ie create a DataLoader with CuArrays instead of Array?

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -0,0 +1,131 @@
...
+    
+    ytrain, ytest = onehotbatch(ytrain, 0:9), onehotbatch(ytest, 0:9)
+
+    train_loader = DataLoader((xtrain, ytrain), batchsize=batchsize, shuffle=true)
+    test_loader = DataLoader((xtest, ytest), batchsize=batchsize)
+
+    return train_loader, test_loader
+end
+
+function ℓ(dataloader, model, device)
```
DhairyaLGandhi, line 54: Beat to call this `loss` for simplicity. Rather than taking in `device` could we not generically use `gpu`? It's free if there is no GPU available and there's little reason to not take the free speed otherwise.

DhairyaLGandhi:
You're welcome to 😉, but removing the model specific env would be appreciated.

SimonCoste:
No problem, but actually all the other vision examples have the env files...

vision/convmixer_cifar10/convmixer.jl
ToucheSir: I'm not sure if all GPUs people might want to try this out on would be able to support the model and all of CIFAR-10. One approach you could try is wrapping the dataloader in [`CUDA.CuIterator`](https://cuda.juliagpu.org/stable/usage/memory/#Batching-iterator), which handles this kind of memory management more efficienty.

DhairyaLGandhi: That would be roughly the same as the current version. Cifar isn't a very heavy dataset, so worse case might be tweaking the batch size.

DhairyaLGandhi:
That's right, but those are harder to maintain as we have seen. A single env is easier to maintain.

vision/convmixer_cifar10/convmixer.jl
ToucheSir: I had a look, and of the [two](https://github.com/FluxML/model-zoo/blob/master/vision/vgg_cifar10/vgg_cifar10.jl) [other](https://github.com/FluxML/model-zoo/blob/master/tutorials/60-minute-blitz/60-minute-blitz.jl) models using CIFAR-10, one uploads everything to GPU and the other works in batches. So as long as VRAM usage stays under, say, 4-6GB with the default hyperparams here, I'm fine with this.

SimonCoste:
I removed the env files and slightly reworked the loss function. Now we have a 
```julia
function create_loss_function(dataloader, device)

    function loss(model)
        # computes the loss and accuracy of the model, on the dataloader, 
        # and on the specified device. 
    end
    return loss
   
end
```

vision/convmixer_cifar10/convmixer.jl
SimonCoste: Maybe, for CIFAR10, we could have a dataloader with `CuArrays`, but this is clearly not doable for any other heavier model. AFAIK, it's really unusual (and often, impossible) to load a whole training dataset to GPU, no?

CarloLucibello:
Thanks for this contribution!

CarloLucibello:
We recently moved from a global Project/Manifest to local ones for maintainability reasons
(e.g. carelessly updating a manifest impacting all scripts is not safe). We should not go back, so I'll add a local manifest/project for this script in another PR (or @SimonCoste can do it if he'd like to)

SimonCoste:
Hi, thanks. I had put the manifest/project files in the first commit but then removed them. Can you retrieve them ?

CarloLucibello:
> Hi, thanks. I had put the manifest/project files in the first commit but then removed them. Can you retrieve them ?

yes. 

I just noticed that the model is from a paper still under review. This is not ideal. Not too bad either, the script is very nice and clean and this repo serves to show examples for educational purpose and for some quick copy/paste/adapt

ToucheSir:
For context, ConvMixer has kicked up a lot of discussion and been included in well-known model libraries too. I would say there is a threshold where something that just has a pre-print becomes well-regarded enough to stand in for a formal publication. What that threshold is I'm not sure.

SimonCoste:
Hi, to be honest, I don't think the Convmixer architecture should be included in model libraries (like MetalHead), at least for the moment - I included it here because I felt ModelZoo is more a showcase of simple examples, training loops, etc.

PR #338, opened 2022-02-21
Initial PR just for Titanic Model

arcAman07:
Initially as I was told, this PR contains the function which fetches the data from the MLDatasets Library, divides them into training and testing set so that it can be used to create the ML model using Flux.jl

Things to be added: Updated ReadMe, Toml file consisting of the dependency and updating docs.

### DIFF
--- a/other/Titanic/titanic.jl
+++ b/other/Titanic/titanic.jl
@@ -1,0 +1,21 @@
+using MLDatasets:Titanic
+
+function get_processed_data(args)
+    labels = Titanic.targets()
+    features = Titanic.features()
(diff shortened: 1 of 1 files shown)


darsnack:
While there is nothing wrong with what's already in the PR, this will need all the "things to add" like a README, toml, complete training script, etc. before it is ready for review.

You do not need to close this PR, just keep pushing more commits.

arcAman07:
Cool will keep adding the commits, was told the initial PR will be for just processing the data which will be ready to use by the Flux library

PR #342, opened 2022-03-20
Fix a dead link in README

Saransh-cpp:
The particular docs section has been moved to somewhere else. Fixed it!


ToucheSir (approved):
Thanks!

PR #345, opened 2022-04-07
add simple example of time series regression with LSTM

mcreel:
This was discussed on discourse. This first file shows how to use LSTM to forecast data from a simple AR1 model. The neural net works on par with the maximum likelihood estimator.


mcreel:
I'm closing this for now, the PR is premature, I think.

ToucheSir:
If you just need some time to add tweaks or revisions, feel free to mark as a draft instead.

mcreel:
Thanks, but I realized I'm using the LSTM layer incorrectly. I will work on it, and make another PR when I get it right,

PR #348, opened 2022-04-12
added explicit 'flatten' function import

### LINKED ISSUE #346: flatten is no longer exported from Flux
So we need to import `flatten` explicitly to use it. This affects some models, e.g. the convolutional LeNet5 model

LF-Flores:
Fixes #346


DhairyaLGandhi:
Perfect, thanks!

PR #349, opened 2022-04-19
MNIST Score-Based Generative Modeling (Diffusion Model)

jakee417:
Porting [Generative Modeling by Estimating Gradients of the Data Distribution](https://yang-song.github.io/blog/2021/score/) to Julia!

I did have [some trouble](https://github.com/FluxML/model-zoo/compare/master...jakee417:master#diff-c28baf7dd56d1666dbc79d43438f348263f9112f015b865e280ca207724933a9R147) with `ConvTranspose` recreating PyTorch's `ConvTranspose2d` and the `output_padding` kwarg.

Leaving [Probability Flow ODE](https://yang-song.github.io/blog/2021/score/#probability-flow-ode) as a `TODO` since I would like to involve [DifferentialEquations.jl](https://diffeq.sciml.ai/stable/).


vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+# https://arxiv.org/pdf/2011.13456.pdf
+
+using MLDatasets
+using Flux
+using Flux: @functor, chunk, params
+using Flux.Data: DataLoader
+using Parameters: @with_kw
+using BSON
+using CUDA
+using DrWatson: struct2dict
```
DhairyaLGandhi, line 15: If we write a simple utility, we can avoid a package dependency here. A separate tutorial showing project tracking for Flux with DrWatson would be an awesome addition!

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+In our case, MNIST comes in as (28, 28, 1, batch). \n
+
+# References
+paper-  https://arxiv.org/abs/1505.04597 \n
+Conv-   https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Conv \n
+TConv-  https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.ConvTranspose \n
+GNorm-  https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.GroupNorm \n
+Flux-   https://fluxml.ai/Flux.jl/stable/models/advanced/#Custom-Model-Example
+"""
+struct UNet
```
DhairyaLGandhi, line 74: Could we reuse UNet.jl here? The projection can be added like `Chain(GaussProj(t), UNet())`. It should simplify the pr a lot!

DhairyaLGandhi (commented):
Diffusion models are awesome! Thanks for the PR, it seems very interesting. I've added a couple thoughts related to making the PR easier for you. What do you think of using UNet.jl? We can make it take activation as an argument, but we also have some changes planned.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Yes! I think thats a fantastic idea. Is [this](https://github.com/DhairyaLGandhi/UNet.jl/blob/master/src/model.jl) what you are referring to?

If so, I do not think it would be as simple as `Chain(GaussProj(t), UNet())`, since we also add in `dense(embed)` after each block of the `UNet`. An example of that is [here](https://github.com/FluxML/model-zoo/pull/349/files#diff-c28baf7dd56d1666dbc79d43438f348263f9112f015b865e280ca207724933a9R217). However we can adapt my [`function (unet::UNet)(x, t)`](https://github.com/FluxML/model-zoo/pull/349/files#diff-c28baf7dd56d1666dbc79d43438f348263f9112f015b865e280ca207724933a9R211) to your [`function (u::Unet)(x::AbstractArray)`](https://github.com/DhairyaLGandhi/UNet.jl/blob/master/src/model.jl#L72) to condition on the positional encodings. Additionally, we would need to add the `Dense` layers to [`function Unet(channels::Int = 1, labels::Int = channels)`](https://github.com/DhairyaLGandhi/UNet.jl/blob/master/src/model.jl#L49) as well and ensure the dimensions still work out correctly (which was already a little tricky to do in my current PR 😂). And as you point out, we would need `swish` activations. The `GaussFourierProject(t)` and rescaling by `marginal_prob_std` can happen independently of the `UNet` so no need to include them in this conversation.

Given this, do you think we would really be simplifying the current PR 😅? I feel like adapting my existing PR to `UNet.jl` is the same amount of code as defining it directly in `diffusion_mnist`.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: What is meant by project tracking? Would the simple dependency just be converting a struct to a dict so that we can serialize with `BSON`? I was trying to keep my already long PR as short as possible, but if you think adding a separate utility to accomplish this is of value, I would be happy to take a look 😃

vision/diffusion_mnist/diffusion_mnist.jl
CarloLucibello: It's good for models' implementations to live in this repo. 
At some point we could introduce examples using models from mainstream repos such as Metalhead.jl, but not in this PR.

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+
+# References:
+https://www.tensorflow.org/api_docs/python/tf/expand_dims
+"""
+expand_dims(x::AbstractVecOrMat, dims::Int=2) = reshape(x, (ntuple(i -> 1, dims)..., size(x)...))
+
+"""
+Helper function that reverses the order of dimensions.
+"""
+reverse_dims(x) = permutedims(x, reverse(ntuple(i -> i, length(size(x)))))
```
CarloLucibello, line 194: `expand_dims` and `reverse_dims` could be nice additions to MLUtils.jl

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+        Dense(embed_dim, channels[2]),
+        GroupNorm(channels[2], 32),
+        ConvTranspose((3, 3), channels[2] + channels[2] => channels[1], pad=(0, -1, 0, -1), stride=2, bias=false),
+        Dense(embed_dim, channels[1]),
+        GroupNorm(channels[1], 32),
+        ConvTranspose((3, 3), channels[1] + channels[1] => 1, stride=1, bias=false),
+        # Scaling Factor
+        marginal_prob_std
+    )
+end
```
CarloLucibello, line 178: Just a tangential comment, but I wonder if we could avoid some typing in models' creation using a pattern like this:
```julia
struct UNet
  layers::NamedTuple
end

function UNet(marginal_prob_std, channels=[32, 64, 128, 256], embed_dim=256, scale=30.0f0)
  return UNet((proj = GaussianFourierProjection(embed_dim, scale),
                       dense1 = Dense(embed_dim, embed_dim),
                       ...))
end        
```

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+
+# References:
+http://www.iro.umontreal.ca/~vincentp/Publications/smdae_techreport.pdf \n
+https://yang-song.github.io/blog/2021/score/#estimating-the-reverse-sde-with-score-based-models-and-score-matching \n
+https://yang-song.github.io/blog/2019/ssm/
+"""
+function model_loss(model, x, device, ϵ=1.0f-5)
+    batch_size = size(x)[end]
+    # (batch) of random times to approximate 𝔼[⋅] wrt. 𝘪 ∼ 𝒰(0, 𝘛)
+    random_t = rand(Float32, batch_size) .* (1.0f0 - ϵ) .+ ϵ |> device
```
CarloLucibello, line 277: what is this -eps + eps for?

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+    # (batch) of perturbations to approximate 𝔼[⋅] wrt. 𝘹(0) ∼ 𝒫₀(𝘹)
+    z = randn(Float32, size(x)) |> device
```
CarloLucibello, line 279: z = randn!(similar(x))

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+    @info "Start Training, total $(args.epochs) epochs"
+    for epoch = 1:args.epochs
+        @info "Epoch $(epoch)"
+        progress = Progress(length(loader))
+
+        for (x, _) in loader
+            loss, back = Flux.pullback(ps) do
+                model_loss(unet, x |> device, device)
+            end
+            grad = back(1.0f0)
```
CarloLucibello, line 360:
```suggestion
            loss, grad = Flux.withgradient(ps) do
                model_loss(unet, x |> device, device)
            end
```

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+
+Notes:
+    ```julia
+    @assert size(a) == (28, 28, 1, 32, 32)
+    @assert size(b) == (32, 32)
+    o = reverse_dims(a) .+ reverse_dims(b)' |> reverse_dims
+    size(o)  # (28, 28, 1, 32, 32)
+    ```
+    Is used to broadcast an operation from b to a 
+    without the cost of a copy (using expand_dims for example)
```
CarloLucibello, line 207: `expand_dims` is a reshape, so no copy involved, while `reverse_dims(a)` is an expensive operation

CarloLucibello, line 207: `expand_dims` is a reshape, so no copy involved, while `reverse_dims(a)` is an expensive operation

vision/diffusion_mnist/diffusion_plot.jl
```diff
@@ -0,0 +1,121 @@
...
+
+# References
+https://yang-song.github.io/blog/2021/score/#how-to-solve-the-reverse-sde
+"""
+function Euler_Maruyama_sampler(model, init_x, time_steps, Δt, device)
+    x = mean_x = init_x
+    progress = Progress(length(time_steps))
+    @info "Start Euler-Maruyama Sampling"
+    for time_step in time_steps
+        batch_time_step = ones(Float32, size(init_x)[end]) .* time_step |> device
```
CarloLucibello, line 53: batch_time_step = fill!(similar(init_x, size(init_x)[end]), 1) .* time_step

CarloLucibello:
This is a great addition, thanks! And looking very good already. 

Let's remove the saved bson models, and remove them from the git history (e.g. with a `git rebase -i HEAD~10`)

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: [Here](https://colab.research.google.com/drive/120kYYBOVa1i0TD85RjlEkFjaWDxSFUx3#scrollTo=zOsoqPdXHuL5&line=16&uniqifier=1) is the original where they claim "eps: A tolerance value for numerical stability."

Intuitively, we are taking an expectation over `random_t`, where `random_t = 0` corresponds to the unknown data distribution (in this case the distribution of MNIST) and `random_t = 1` corresponds to our prior (which is known analytically to be jointly Gaussian). My guess is that we only need to learn the transition times `0 < random_t < 1` because the sampling process will take in `random_t = 1` as an input and produce values at `random_t = 0`.

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,385 @@
...
+    which is not allowed (mutating vector) by Zygote in 
+    a custom model.
+"""
+function (unet::UNet)(x, t)
+    # Embedding
+    embed = unet.gaussfourierproj(t)
+    embed = swish.(unet.linear(embed))
+    # Encoder
+    h1 = unet.conv1(x)
+    h1 = reverse_dims(h1) .+ unet.dense1(embed)' |> reverse_dims
```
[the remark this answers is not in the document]
jakee417: No, this will work fine! I will go ahead and get rid of the `reverse_dims` sillyness, as this is unneeded.

Full story, I originally had:
``` julia
h1 .+= expand_dims(unet.dense1(embed), 2)
``` 
and getting
```julia
ERROR: Mutating arrays is not supported -- called copyto!(::Array{Float32, 4}, _...)
```
which I **incorrectly** deduced `expand_dims` being a copy and the source of the issue.

I thought it was odd reshape was doing a copy... now I understand its the `.+=` operator 🤦‍♂️

vision/diffusion_mnist/diffusion_mnist.jl
CarloLucibello: this shift and rescale makes t ~ Unif(eps, 1),  maybe there are some problems with t exactly 0

vision/diffusion_mnist/diffusion_mnist.jl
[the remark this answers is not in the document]
jakee417: `h_x = reverse_dims(h_x) .+ unet.dense_x(embed)' |> reverse_dims` (throughout entire network):
```julia
julia>     @time [unet_test(x_test, t_test) for _ in 1:500];
 64.377195 seconds (540.61 k allocations: 41.847 GiB, 0.88% gc time, 0.05% compilation time)
```

`h_x = h_x .+ expand_dims(unet.dense_x(embed), 2)` (throughout entire network):
```julia
julia>     @time [unet_test(x_test, t_test) for i in 1:500];
 64.515426 seconds (529.79 k allocations: 33.513 GiB, 0.27% gc time, 0.04% compilation time)
```
So your suspicion was correct, it is less efficient memory-wise **and** way less pretty style-wise.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Its possible, but requires
```julia
model.layers.dense1(x)
```
which is a bit clunky. I think its worth the trade since the layers are now tied to their name 😁

vision/diffusion_mnist/diffusion_mnist.jl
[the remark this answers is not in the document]
ToucheSir: FYI that https://github.com/JuliaCI/BenchmarkTools.jl will make your life much easier for this kind of timing ;)

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Much clearer, thanks!

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+
+"""
+Projection of Gaussian Noise onto a time vector.
+
+# Notes
+This layer will help embed our random times onto the frequency domain. \n
+W is not trainable and is sampled once upon construction - see assertions below.
+
+# References
+paper-  https://arxiv.org/abs/2006.10739 \n
```
ToucheSir, line 29:
```suggestion
paper-  https://arxiv.org/abs/2006.10739
```

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+"""
+function GaussianFourierProjection(embed_dim, scale)
+    # Instantiate W once
+    W = randn(Float32, embed_dim ÷ 2) .* scale
+    # Return a function that always references the same W
+    function GaussFourierProject(t)
+        t_proj = t' .* W * Float32(2π)
+        [sin.(t_proj); cos.(t_proj)]
+    end
+end
```
ToucheSir, line 40: If `W` is learnable, this should be a struct type. Ref. https://fluxml.ai/Flux.jl/stable/models/advanced/.

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+    # 𝘚₀(𝘹(𝘵), 𝘵)
+    score = model(perturbed_x, random_t)
+    # mean over batches
+    sum(
+        # L₂ norm over WHC dimensions
+        sum(
+            (score .* std + z) .^ 2,
+            dims=(1, 2, 3)
+        )
+    ) / batch_size
```
ToucheSir, line 230: The outer sum can just be a `mean`, no?

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+        ########################################################################
+        tconv3=ConvTranspose((3, 3), channels[3] + channels[3] => channels[2], pad=(0, -1, 0, -1), stride=2, bias=false),
+        dense6=Dense(embed_dim, channels[2]),
+        tgnorm3=GroupNorm(channels[2], 32),
+        tconv2=ConvTranspose((3, 3), channels[2] + channels[2] => channels[1], pad=(0, -1, 0, -1), stride=2, bias=false),
+        dense7=Dense(embed_dim, channels[1]),
+        tgnorm2=GroupNorm(channels[1], 32),
+        tconv1=ConvTranspose((3, 3), channels[1] + channels[1] => 1, stride=1, bias=false),
+        # Scaling Factor
+        marginal_prob_std=marginal_prob_std
```
ToucheSir, line 131: Unless the idea is to make this configurable, a stateless function need not be a sublayer. Calling it directly in the forward pass would also be faster.

vision/diffusion_mnist/diffusion_plot.jl
```diff
@@ -0,0 +1,121 @@
+include("diffusion_mnist.jl")
+
+using Images
+
+"""
+Helper function yielding the diffusion coefficient from a SDE.
+"""
+function diffusion_coeff(t, sigma=25.0f0)
+    sigma .^ t
+end
```
ToucheSir, line 10:
```suggestion
diffusion_coeff(t, sigma=convert(eltype(t), 25.0f0)) = sigma .^ t
```

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+    # Training
+    train_steps = 0
+    @info "Start Training, total $(args.epochs) epochs"
+    for epoch = 1:args.epochs
+        @info "Epoch $(epoch)"
+        progress = Progress(length(loader))
+
+        for (x, _) in loader
+            loss, grad = Flux.withgradient(ps) do
+                model_loss(unet, x |> device, device)
```
ToucheSir, line 304: It's always better to move `x` to cpu/gpu _outside_ of the gradient callback unless absolutely necessary.

ToucheSir: Which also implies that `model_loss` shouldn't (have) to take a `device` argument either. Happy to elaborate once you've updated to use `similar` for most bits :)

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+"""
+function (unet::UNet)(x, t)
+    # Embedding
+    embed = unet.layers.gaussfourierproj(t)
+    embed = swish.(unet.layers.linear(embed))
+    # Encoder
+    h1 = unet.layers.conv1(x)
+    h1 = h1 .+ expand_dims(unet.layers.dense1(embed), 2)
+    h1 = unet.layers.gnorm1(h1)
+    h1 = swish.(h1)
```
ToucheSir, line 157: [Dense](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Dense) and [GroupNorm](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.GroupNorm) both take an activation function, so adding the external broadcast is unnecessary (and likely less efficient too).

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+
+We use properties of SDEs to analytically solve for the stddev
+at time t conditioned on the data distribution. \n
+
+We will be using this all over the codebase for computing our model's loss,
+scaling our network output, and even sampling new images!
+"""
+function marginal_prob_std(t, sigma=25.0f0)
+    sqrt.((sigma .^ (2t) .- 1.0f0) ./ 2.0f0 ./ log(sigma))
+end
```
ToucheSir, line 57: This could probably be a one-liner. More generally, I'm surprised there isn't a pre-made function for this in some existing library (or is there?)

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,330 @@
...
+# Notes
+Images stored in WHCN (width, height, channels, batch) order. \n
+In our case, MNIST comes in as (28, 28, 1, batch). \n
+
+# References
+paper-  https://arxiv.org/abs/1505.04597 \n
+Conv-   https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Conv \n
+TConv-  https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.ConvTranspose \n
+GNorm-  https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.GroupNorm \n
+Flux-   https://fluxml.ai/Flux.jl/stable/models/advanced/#Custom-Model-Example
```
ToucheSir, line 71: I don't think references to Flux layers need to be in the docstrings. External source material is sufficient.
```suggestion
```

vision/diffusion_mnist/diffusion_plot.jl
```diff
@@ -0,0 +1,121 @@
...
+    progress = Progress(length(time_steps))
+    @info "Start PC Sampling"
+    for time_step in time_steps
+        batch_time_step = ones(Float32, size(init_x)[end]) .* time_step |> device
+        # Corrector step (Langevin MCMC)
+        grad = model(x, batch_time_step)
+        num_pixels = prod(size(grad)[1:end-1])
+        grad_batch_vector = reshape(grad, (size(grad)[end], num_pixels))
+        grad_norm = sqrt.(sum(abs2, grad_batch_vector, dims=2))
+        grad_norm = sum(grad_norm) / length(grad_norm)
```
ToucheSir, line 79:
```suggestion
        grad_norm = mean(sqrt, sum(abs2, grad_batch_vector, dims=2))
```

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: perhaps they could have done `t * (1 - 2 * eps) + eps` so that `t' ~ Unif(eps, 1 - eps)`?

Either way you are correct:
``` julia
h ./ expand_dims(unet.layers.marginal_prob_std(t), 3)
```
results in `h ./ 0` for some component when `t=0` since `marginal_prob_std(0) = 0`. So probably just a little fudge factor 😊

vision/diffusion_mnist/diffusion_mnist.jl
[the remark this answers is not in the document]
DhairyaLGandhi: For benchmarks taking >3s, its usually the case that `@time` is sufficient

vision/diffusion_mnist/diffusion_mnist.jl
DhairyaLGandhi: Yeah, we can make it so the original repo takes some up some of the changes we need for this model. I am open to that.

vision/diffusion_mnist/diffusion_mnist.jl
DhairyaLGandhi: Ideally, the device on which some arrays live should already be defined before any function evaluation. That is to say that ideally we shouldn't need to have more than a couple invocations to `cpu/gpu` ahead of training.

vision/diffusion_mnist/diffusion_mnist.jl
DhairyaLGandhi: Was about to say the same. Also the reason I mentioned taking activation as an arg in UNet.jl

vision/diffusion_mnist/diffusion_mnist.jl
DhairyaLGandhi: To me the implementation looks like `random_t` needs to be coming from the data loader

vision/diffusion_mnist/diffusion_plot.jl
jakee417: Like this, because now we don't have to pass around `device`.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Yes, `W` is purposely not learnable.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: I couldn't mind a mean without using another import... figured `sum() / length()` was easy enough 🤷‍♂️

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Yeah I think I was just blindly following the PyTorch tutorial. Removing the sublayer also means we don't need to pass it to `UNet()` upon construction 😎

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Yep! Makes the forward pass alot less bloated!

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Not to my knowledge... do one liner functions still have docstrings? I feel like some level of explanation is needed here to motivate the operation. I could also move as a normal comment if one-liner is much preferred

jakee417: image

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Sorry, why can we not construct `random_t` on the fly? They are just randomly sampled indexes into time, they don't exist as a datasource outside the loss function.

vision/diffusion_mnist/diffusion_mnist.jl
ToucheSir: `Statistics.mean` is part of the stdlib and should be favoured for many reasons: performance, semantic correctness, terseness (we try to avoid breaking nested expressions like this over too many lines), best practices (many new users don't know it exists). I would not think twice about using it :)

vision/diffusion_mnist/diffusion_mnist.jl
ToucheSir: Yes, docstrings work exactly the same way.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Would this be the fix?
```julia
      x = x |> device
      for (x, _) in loader
          loss, grad = Flux.withgradient(ps) do
              model_loss(unet, x, device)
```

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: I would be happy to contribute `expand_dims`! You still think there is value in `reverse_dims`? I thought it was kinda hack.

Could you just point me to the file in MLUtils.jl this should live?

vision/diffusion_mnist/diffusion_mnist.jl
ToucheSir: Did you mean:
```julia
for (x, _) in loader
  x = device(x)
  ...
```
?

vision/diffusion_mnist/Project.toml
```diff
@@ -0,0 +1,13 @@
...
+DrWatson = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
+Images = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+MLDatasets = "eb30cadb-4394-5ae3-aed4-317e484a6458"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
+ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+TensorBoardLogger = "899adc3e-224a-11e9-021f-63837185c80f"
```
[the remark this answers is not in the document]
jakee417: I used the `vae_mnist.jl` as a template so I copied these over from its `Project.toml` file:
```
[compat]
CUDA = "2.4"
Flux = "0.11.4, 0.12"
julia = "1.5"
```
Not sure if this is what you were looking for 🤷‍♂️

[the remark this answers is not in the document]
ToucheSir: I would use the latest version of all packages (you can find this via `] st`). You can use the major version for any post-1.0 libraries and `0.X` for others (including Flux itself). Julia should be at least 1.6 since that's the current LTS.

vision/diffusion_mnist/diffusion_mnist.jl
jakee417: Oh, yep! I will add next commit

DhairyaLGandhi: I think the better solution is to do something like 

```julia

X = device(x)
model = device(x)

loader = DataLoader(X)
for in loader
...
```

vision/diffusion_mnist/diffusion_mnist.jl
ToucheSir:
```suggestion
    random_t = rand!(similar(x, batch_size)) .* (1.0f0 - ϵ) .+ ϵ
```
Generating on the fly seems fine, though you can get away with not using `device`. AFAICT this is the only remaining use in `model_loss`, so the arg can be removed completely.

vision/diffusion_mnist/diffusion_mnist.jl
```diff
@@ -0,0 +1,316 @@
...
+    perturbed_x = x + z .* std
+    # 𝘚₀(𝘹(𝘵), 𝘵)
+    score = model(perturbed_x, random_t)
+    # mean over batches
+    mean(
+        # L₂ norm over WHC dimensions
+        sum(
+            (score .* std + z) .^ 2,
+            dims=(1, 2, 3)
+        )
```
ToucheSir, line 214:
```suggestion
        sum((score .* std + z) .^ 2; dims=1:(ndims(x) - 1))
```
Minor formatting nit.

vision/diffusion_mnist/diffusion_test.jl
```diff
@@ -0,0 +1,40 @@
...
+
+    @assert expand_dims(ones(Float32, 32), 3) |> size == (1, 1, 1, 32)
+
+    unet_test = UNet()
+    x_test = randn(Float32, (28, 28, 1, 32))
+    t_test = rand(Float32, 32)
+    score_test = unet_test(x_test, t_test)
+    @assert score_test |> size == (28, 28, 1, 32)
+    @assert typeof(score_test) == Array{Float32,4}
+    @time [unet_test(x_test, t_test) for _ in 1:10];
```
ToucheSir, line 25:
```suggestion
```

vision/diffusion_mnist/diffusion_plot.jl
```diff
@@ -0,0 +1,118 @@
...
+    Gray.(
+        permutedims(
+            vcat(
+                reshape.(
+                    chunk(x |> cpu, y_size), 28, :
+                )...
+            ),
+            (2, 1)
+        )
+    )
```
ToucheSir, line 23: The expression here can be split up into 2-3 lines without creating a pyramid. Though I feel there must be an easier way to do this conversion too...

vision/diffusion_mnist/diffusion_plot.jl
```diff
@@ -0,0 +1,118 @@
...
+    progress = Progress(length(time_steps))
+    @info "Start Euler-Maruyama Sampling"
+    for time_step in time_steps
+        batch_time_step = fill!(similar(init_x, size(init_x)[end]), 1) .* time_step
+        g = diffusion_coeff(batch_time_step)
+        mean_x = x .+ expand_dims((g .^ 2), 3) .* model(x, batch_time_step) .* Δt
+        x = mean_x + sqrt(Δt) * expand_dims(g, 3) .* randn(Float32, size(x))
+        next!(progress; showvalues=[(:time_step, time_step)])
+    end
+    return mean_x
```
ToucheSir, line 57:
```suggestion
    @info "Start Euler-Maruyama Sampling"
    @showprogress for time_step in 1:length(time_steps)
        batch_time_step = fill!(similar(init_x, size(init_x)[end]), 1) .* time_step
        g = diffusion_coeff(batch_time_step)
        mean_x = x .+ expand_dims(g, 3) .^ 2 .* model(x, batch_time_step) .* Δt
        x = mean_x .+ sqrt(Δt) .* expand_dims(g, 3) .* randn(Float32, size(x))
    end
    return mean_x
```
This can be simplified further by using `@showprogress "Start Euler-Maruyama Sampling" for ...`.

CarloLucibello:
Should we ignore the ConTranspose problem for the time being? Is there a corresponding issue in Flux?

jakee417:
> Should we ignore the ConTranspose problem for the time being? Is there a corresponding issue in Flux?

Yep! Removed and adding here just to track.

Relevant issues are [here](https://github.com/FluxML/Flux.jl/issues/1319) and [here](https://github.com/FluxML/Flux.jl/issues/829).

``` julia
        ########################################################################
        # FIXME: Julia does not offer a `output_padding` kwarg such as in:
        # https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html#convtranspose2d
        #
        # A fix is referenced here: https://github.com/FluxML/Flux.jl/issues/1319
        # and here: https://github.com/FluxML/Flux.jl/issues/829
        # But padding with `pad=SamePad()` or `pad=(0, 1, 0, 1)` yields (10, 10, 64, 3) 
        # which is still incorrect.
        # 
        # Based off relationship 14 of https://arxiv.org/pdf/1603.07285.pdf:
        # 𝘰' = 𝘴(𝘪' - 1) + 𝘢 + 𝘬 - 2𝘱
        # Set 𝘰' = 12, 𝘴 = 2, 𝘪' = 5, 𝘬 = 3, 𝘱 = 0, then
        # 12 = 2(5 - 1) + a + 3 - 2(0) = 11 + a => a = 1
        # However, no Flux API that I could find exposes the argument a for the user...
        #
        # A (incorrect) hack to get the shapes to work is just to pad by (-1, 0, -1, 0):
        # julia> randn(5, 5, 256, 3) |> ConvTranspose((3, 3), 256 => 64, stride=2, pad=(0, -1, 0, -1)) |> size
        # (12, 12, 64, 3)
        #
        # Which is the correct shape, but seems suspicious (negative padding??).
        # Why is passing a negative padding even allowed in the first place? 😕
        ########################################################################
```

vision/diffusion_mnist/diffusion_plot.jl
jakee417: collapsed the pyramid. I guess extra typing is avoided in Julia? I feel the pyramid helps readability

jakee417:
Running a 30-epoch test just to confirm edits didn't break performance, but code is looking nice and concise now! Thanks all for the input 😊

CarloLucibello (approved):

CarloLucibello:
Seems good to merge whenever you are ready

vision/diffusion_mnist/diffusion_plot.jl
jakee417: Had to switch
```julia
 @showprogress for time_step in 1:length(time_steps)
```

to 
```julia
 @showprogress for time_step in time_steps
```
But other than that, this works fine

jakee417:
Fixed a very small bug and got these results from 30 epochs:
**Sampled Noise**
sampled_noise
**EM Sampler**
em_images
**PC Sampler**
pc_images

ToucheSir:
Thanks for bearing with us @jakee417, this is a great addition to the zoo!

PR #352, opened 2022-05-01
Add example of autoregressive process modelling

JLDC:
Provides a short example of how to create a recurrent model for univariate time series data based on an AR process.


other/autoregressive-process/utils.jl
```diff
@@ -0,0 +1,42 @@
...
+function generate_process(ϕ::AbstractVector{Float32}, s::Int)
+    @assert s > 0 "s must be positive"
+    # Generate white noise
+    ϵ = randn(Float32, s)
+    # Initialize time series
+    X = zeros(Float32, s)
+    p = length(ϕ)
+    X[1] = ϵ[1]
+    # Reverse the order of the coefficients for multiplication later on
+    reverse!(ϕ)
```
ToucheSir, line 14:
```suggestion
    ϕ = reverse(ϕ) 
```
For safety. The overhead of copying a 3-vector should be negligible.

other/autoregressive-process/utils.jl
```diff
@@ -0,0 +1,42 @@
...
+            X[t] = sum(X[t-p:t-1] .* ϕ) + ϵ[t]
+        end
+    end
+    X
+end
+
+# Create batches of a time series `X` by splitting the series into
+# sequences of length `s`. Each new sequence is shifted by `r` steps.
+# When s == r,  the series is split into non-overlapping batches.
+function batch_timeseries(X, s::Int, r::Int)
```
ToucheSir, line 29: Is there no existing functionality for this in https://github.com/JuliaML/MLUtils.jl? If not, perhaps there should be.

other/autoregressive-process/model.jl
```diff
@@ -0,0 +1,91 @@
...
+    # Training loop
+    for i ∈ 1:args.epochs
+        Flux.reset!(m) # Reset hidden state of the recurrent model
+        # Compute the gradients of the loss function
+        ∇ = gradient(θ) do
+            loss(Xtrain, ytrain)
+        end
+        Flux.update!(opt, θ, ∇) # Update model parameters
+        if args.verbose && i % 10 == 0 # Log results every 10 epochs
+            Flux.testmode!(m)
```
ToucheSir, line 75:
```suggestion
```
This is a no-op since the model doesn't have any normalization layers (specifically dropout, batchnorm or groupnorm).

other/autoregressive-process/model.jl
```diff
@@ -0,0 +1,91 @@
...
+            # Compute loss on train and test set for logging (important: the model must be reset!)
+            Flux.reset!(m)
+            train_loss = loss(Xtrain, ytrain)
+            Flux.reset!(m)
+            test_loss = loss(Xtest, ytest)
+            @info "Epoch $i / $(args.epochs), train loss: $(round(train_loss, digits=3)) | test loss: $(round(test_loss, digits=3))"
+            Flux.trainmode!(m) # Put the mode back into train mode
```
ToucheSir, line 82:
```suggestion
```
Same as above.

other/autoregressive-process/utils.jl
```diff
@@ -0,0 +1,42 @@
...
+    # Reverse the order of the coefficients for multiplication later on
+    reverse!(ϕ) 
+    # Compute values iteratively
+    for t ∈ 2:s
+        if t ≤ p
+            X[t] = sum(X[1:t-1] .* ϕ[1:t-1]) + ϵ[t]
+        else
+            X[t] = sum(X[t-p:t-1] .* ϕ) + ϵ[t]
+        end
+    end
```
ToucheSir, line 22: This can definitely be simplified, I'm not exactly sure of how much though. Can the loop be expressed as a convolution?

other/autoregressive-process/utils.jl
JLDC: I believe there is not, this is also something I want to add to https://fluxml.ai/Flux.jl/stable/models/recurrence/#Working-with-sequences but I was unsure on whether it was wanted/useful. I also have a similar function to reverse the batched time series into a tabular format. If there is already a functionality, I'm happy to learn about it.

ToucheSir (commented):
This looks great! Mostly style and cleanup comments.

other/autoregressive-process/utils.jl
JLDC: I made the following change
```julia
# Fill first p observations
for t ∈ 1:p-1
    X[t+1] = X[1:t]'ϕ[1:t] + ϵ[t+1]
end
# Compute values iteratively
for t ∈ p+1:s
    X[t] = X[t-p:t-1]'ϕ + ϵ[t]
end
```
which is a bit faster. I don't think I can use a `mapreduce` here but perhaps someone more knowledgeable can enlighten me.

ToucheSir (approved):
Ok, I think this is good to go. Any final tweaks you want to add or comments from the floor?

ToucheSir:
Thanks again, @JLDC!

PR #362, opened 2022-07-14
Update conv_mnist.jl

alex-s-gardner:
DataLoader type no longer includes nobs, therefore:

replace: 
    @info "Dataset MNIST: $(train_loader.nobs) train and $(test_loader.nobs) test examples"
with:
@info "Dataset MNIST: $(size(train_loader.data[1],4)) train and $(size(test_loader.data[1],4)) test examples"


ToucheSir:
`nobs` has now become https://juliaml.github.io/MLUtils.jl/dev/api/#MLUtils.numobs, so you'll want to use that instead.

PR #369, opened 2022-11-02
Updated data source and loss function

luboshanus:
Updated data loader to be functional with `Flux` and changed `loss` function for the one of `Flux.Losses`.

- [ ] no gpu support
- [x] uses deprecated Flux.Data dataset
- [ ] could use MLDataPatterns.splitobs
- [x] uses deprecated logitcrossentropy

### PR Checklist

- [ ] Tests are added
- [ ] Documentation, if applicable

### DIFF
--- a/other/iris/iris.jl
+++ b/other/iris/iris.jl
@@ -18,7 +18,7 @@
 # To run this example, we need the following packages:
 
 using Flux
-using Flux: logitcrossentropy, normalise, onecold, onehotbatch
+using Flux: logitcrossentropy, normalise, onecold, onehotbatch, params
 using Statistics: mean
 using Parameters: @with_kw
 
@@ -37,8 +37,8 @@ end
 
 
 function get_processed_data(args)
-    labels = Flux.Data.Iris.labels()
-    features = Flux.Data.Iris.features()
+    labels = Flux.load_iris()[2]
+    features = Flux.load_iris()[1]
 
     ## Subract mean, divide by std dev for normed mean of 0 and std dev of 1.
     normed_features = normalise(features, dims=2)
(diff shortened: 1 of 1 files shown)


other/iris/iris.jl
```diff
@@ -37,8 +37,8 @@ end
 
 
 function get_processed_data(args)
-    labels = Flux.Data.Iris.labels()
-    features = Flux.Data.Iris.features()
+    labels = Flux.load_iris()[2]
+    features = Flux.load_iris()[1]
```
ToucheSir, line 41: Datasets are now just an `] add` away in MLDatasets.jl

other/iris/iris.jl
```diff
@@ -18,7 +18,7 @@
 # To run this example, we need the following packages:
 
 using Flux
-using Flux: logitcrossentropy, normalise, onecold, onehotbatch
+using Flux: logitcrossentropy, normalise, onecold, onehotbatch, params
```
ToucheSir, line 21: I would prefer either both or neither of `logitcrossentropy` and `params` be used qualified or imported. Currently, this imports both but only uses one.

other/iris/iris.jl
```diff
@@ -102,7 +102,7 @@ function train(; kws...)
 	
     ## Define loss function to be used in training
     ## For numerical stability, we use here logitcrossentropy
-    loss(x, y) = logitcrossentropy(model(x), y)
+    loss(x, y) = Flux.Losses.logitcrossentropy(model(x), y)
```
ToucheSir, line 105: See above comment.

ToucheSir (commented):
Thanks. A couple suggestions if you're interested on developing this further.

other/iris/iris.jl
luboshanus: Understand the point, I am sorry for not using correct version of Flux, which still contains the data link.

Related to the above and below, what is the best way to do the edit? 

1) Update Flux, add MLDatasets, use the most updated versions and adjust the code?
2) Or just keep the one edit below, adding.`, params`

other/iris/iris.jl
luboshanus: This I would keep, the below is then not necessary.

Is it what you suggest. Also depends on the comment above, what is the way how to edit/update. :)

PR #371, opened 2022-11-29
Add versions to Readme list

mcabbott:
I checked the project (or manifest) files for the version of Flux on which these models were run. Maybe useful to have up front?

I guess it's an admission that they aren't all going to stay perfectly current. Better to be honest though.

Also adds entries for a few recent PRs which were missing from the Readme's list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,12 +53,13 @@ You can contribute in one of the following ways
 ### Update a model
 
 Each example lists the version of Flux for which it was most recently updated.
-Moving them to the latest is a great way to learn!
-
+Bringing them up to the latest is a great way to learn!
 Flux has a [NEWS page](https://github.com/FluxML/Flux.jl/blob/master/NEWS.md) listing important changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
+<img align="right" width="200px" src="https://github.com/FluxML/Optimisers.jl/raw/master/docs/src/assets/logo.png">
```
ToucheSir, line 1:
```suggestion
<img align="right" width="200px" src="https://fluxml.ai/assets/logo.png">
```

ToucheSir (approved):

README.md
mcabbott: This is the big Flux logo. My thinking was just to add the small one, as on most of the other repositories.

ToucheSir: https://avatars.githubusercontent.com/u/26222520 might be a better permalink then. I'd also be fine with dumping the small version into https://github.com/FluxML/fluxml.github.io/tree/main/_assets so we have a central place for it across the org.

mcabbott: Yes that's probably ideal.

PR #372, opened 2022-12-08
Link to MLJFlux examples?

mcabbott:
Should this package's readme list examples which aren't hosted here, which might be of interest? Seems like a somewhat natural place to look. 

This thread points to MLJFlux:

https://discourse.julialang.org/t/flux-categorical-arrays-roc-curves-confusion-matrices/91282/10


ToucheSir (approved):
Good idea. This only makes the model zoo more useful.

PR #373, opened 2022-12-08
Explicit dcgan mnist

mcabbott:
Last I upgraded a few models to use explicit-mode gradients, Flux 0.13.9. 

They do run, but I did not check with a GPU, nor run them for the full number of epochs.

In several of these, I also
* Removed Parameters.jl, this seems weird as it's built-in, and the whole make-a-struct-for-parameters style here seems over-engineered already. 
* Removed elaborate code for defining `device = gpu` etc. Isn't the point of `gpu` that it warns you then does nothing, if used without a GPU?

(Separate commits, to be easy to remove if people have strong feelings.)

### DIFF
--- a/vision/dcgan_mnist/dcgan_mnist.jl
+++ b/vision/dcgan_mnist/dcgan_mnist.jl
@@ -72,25 +72,21 @@ generator_loss(fake_output) = logitbinarycrossentropy(fake_output, 1)
 function train_discriminator!(gen, dscr, x, opt_dscr, hparams)
     noise = randn!(similar(x, (hparams.latent_dim, hparams.batch_size))) 
     fake_input = gen(noise)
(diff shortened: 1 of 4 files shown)


vision/dcgan_mnist/dcgan_mnist.jl
```diff
@@ -99,31 +94,29 @@ function train(; kws...)
     hparams = HyperParams(; kws...)
 
     if CUDA.has_cuda()
```
mcabbott, line 96: I guess this should also really change to [CUDA.functional](https://cuda.juliagpu.org/stable/api/essentials/#CUDA.functional-Tuple{Bool}):
```suggestion
    if CUDA.functional()
```

CarloLucibello (approved):

PR #374, opened 2022-12-09
Explicit convmixer cifar10

mcabbott:
This previously didn't have a manifest. 

Needs a correction to the translation rule for Optimisers.ClipNorm in Flux.

Also, a simple test showed increasing loss, so perhaps something else is broken?


ToucheSir:
The original README does have some notes about achievable accuracy, but I don't believe anyone verified those numbers. One thing that could help is using the built-in ConvMixer model in Metalhead, because that has undergone basic convergence testing.

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -51,89 +58,75 @@ function get_data(batchsize; dataset = MLDatasets.CIFAR10, idxs = nothing)
...
 
-    ps = params(model)
-    opt = Optimiser(
+    opt = OptimiserChain(
             WeightDecay(1f-3), 
-            ClipNorm(1.0),
-            ADAM(η)
+            ClipNorm(1f0),
+            Adam(η),
             )
```
mcabbott, line 101: This exposes a real error: composing `ClipNorm` after another optimiser means it tries to take `norm(::Broadcasted)` which fails:
```
julia> train(epochs=10, images=128)
ERROR: Scalar indexing is disallowed.
...
Stacktrace:
...
 [12] norm(itr::Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{4}, NTuple{4, Base.OneTo{Int64}}, typeof(+), Tuple{CuArray{Float32, 4, CUDA.Mem.DeviceBuffer}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{4}, Nothing, typeof(*), Tuple{Float32, CuArray{Float32, 4, CUDA.Mem.DeviceBuffer}}}}}, p::Float32)
    @ LinearAlgebra ~/julia-9ded051e9f/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:596
 [13] apply!(o::Optimisers.ClipNorm{Float32}, state::Nothing, x::CuArray{Float32, 4, CUDA.Mem.DeviceBuffer}, dx::Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{4}, NTuple{4, Base.OneTo{Int64}}, typeof(+), Tuple{CuArray{Float32, 4, CUDA.Mem.DeviceBuffer}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{4}, Nothing, typeof(*), Tuple{Float32, CuArray{Float32, 4, CUDA.Mem.DeviceBuffer}}}}})
    @ Optimisers ~/.julia/packages/Optimisers/kPdJV/src/rules.jl:584
```

mcabbott: https://github.com/FluxML/Optimisers.jl/issues/127

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -51,89 +58,75 @@ function get_data(batchsize; dataset = MLDatasets.CIFAR10, idxs = nothing)
...
         if epoch%5==0
-            @info "Epoch $epoch : Train loss = $train_loss || Validation accuracy = $test_acc."
+            @info "Epoch $epoch" train_loss test_acc
         end
-
     end
 
-    model = model |> cpu
-    @save "model.bson" model 
+    @save "model.bson" cpu(model)
```
mcabbott, line 126: Should really be updated to Flux.state too

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -1,139 +1,143 @@
...
-    ps = params(model)
-    opt = Optimiser(
-            WeightDecay(1f-3), 
-            ClipNorm(1.0),
-            ADAM(η)
+    # optimiser and its state:
+    opt = OptimiserChain(
+            WeightDecay(1f-3),  # L2 regularisation
+            ClipNorm(1f0),
+            Adam(3f-4),  # learning rate
```
CarloLucibello, line 103: Current practice would be to use `AdamW` and set L2 regularization there

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -1,139 +1,143 @@
...
+
+    # arrays for logging:
+    train_save = zeros(epochs, 2)
+    test_save = zeros(epochs, 2)
+
+    for epoch in 1:epochs
+        for (x1, y1) in train_loader
+            # move one batch at a time to GPU; gpu(train_loader) would be another way
+            x, y = gpu(x1), gpu(y1)
+            grads = gradient(m -> Flux.logitcrossentropy(m(x), y; agg=sum), model)
```
CarloLucibello, line 115: why `agg=sum`?

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -1,139 +1,143 @@
...
     end
 
-    model = model |> cpu
-    @save "model.bson" model 
-    @save "losses.bson" train_save test_save
+    # save the fully trained model
+    cpu_model = cpu(model)
+    BSON.@save "model.bson" cpu_model
+    BSON.@save "losses.bson" train_save test_save
+    # it's generally more robust to save just the state, like this, but
```
CarloLucibello, line 134: but what? I would remove entirely the model saving with bson in favor of state saving with JLD2

vision/convmixer_cifar10/convmixer.jl
```diff
@@ -1,139 +1,143 @@
...
-        #logging
-        train_loss, train_acc = train_loss_fn(model) |> cpu
-        test_loss, test_acc = test_loss_fn(model) |> cpu
-        train_save[epoch,:] = [train_loss, train_acc]
-        test_save[epoch,:] = [test_loss, test_acc]
+        # logging
+        train_loss, train_acc = loss_and_accuracy(model, train_loader)
+        test_loss, test_acc = loss_and_accuracy(model, test_loader)
+        train_save[epoch, :] = [train_loss, train_acc]
+        test_save[epoch, :] = [test_loss, test_acc]
```
CarloLucibello, line 123: Can we remove this history saving entirely? seems useless

vision/convmixer_cifar10/convmixer.jl
mcabbott: I just left this, as it probably doesn't hurt... no idea why it's there, maybe it was like that in what someone followed? Showing what the keyword is doesn't seem awful

vision/convmixer_cifar10/convmixer.jl
mcabbott: I reverted this because maybe it's not a terrible idea to show some variety of approaches, not repeat exactly the same thing in every model?

The way this model is set up, you'd have to copy-paste the code with the sizes etc. to re-build the model before you can load the state back in. So perhaps leaving this is OK?

vision/convmixer_cifar10/convmixer.jl
mcabbott: It's not a crazy way to collect some numbers, maybe this pattern makes sense to someone instead of others?

vision/convmixer_cifar10/convmixer.jl
mcabbott: Yes, but maybe it's good to show how to chain rules?

PR #375, opened 2022-12-09
Explicit vae mnist

mcabbott:
This previously didn't have a manifest

PR #376, opened 2022-12-09
Explicit spatial transformer

mcabbott:
This previously didn't have a manifest

### DIFF
--- a/vision/spatial_transformer/spatial_transformer.jl
+++ b/vision/spatial_transformer/spatial_transformer.jl
@@ -100,6 +100,8 @@ function train_model(opt, localization_net, classifier, train_loader; epoch=1)
         loss, grads = withgradient(localization_net, classifier) do ln, cl
             model_loss(localization_net, classifier, x, y)
         end
+        # Both the optimiser state `opt` and the gradients match a
+        # tuple of the two networks, so we can `update!` all at once: 
         Flux.update!(opt, (localization_net, classifier), grads)
         losses[i] = loss
         ProgressMeter.next!(progress_tracker; showvalues=[(:loss, loss)])
@@ -174,7 +176,6 @@ classifier =
         Dense(256, 10),
     ) |> dev
 
-# ps = Flux.params(localization_net, classifier)
 ## ====
 # create sampling grid
 const sampling_grid = get_sampling_grid(args[:img_size]...) |> dev
--- a/vision/spatial_transformer/spatial_transformer.jl
+++ b/vision/spatial_transformer/spatial_transformer.jl
@@ -11,7 +11,7 @@ using DrWatson
 using LinearAlgebra, Statistics
 using Flux, Zygote, CUDA
 using Flux: batch, onehotbatch, flatten, unsqueeze
-using Flux.Data: DataLoader
+using Flux: DataLoader
 using MLDatasets
 using Base.Iterators: partition
 using Plots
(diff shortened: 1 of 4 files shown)


vision/spatial_transformer/spatial_transformer.jl
```diff
@@ -174,16 +174,16 @@ classifier =
...
 
-ps = Flux.params(localization_net, classifier)
+# ps = Flux.params(localization_net, classifier)
 ## ====
 # create sampling grid
 const sampling_grid = get_sampling_grid(args[:img_size]...) |> dev
 ## ====
 
-opt = Adam(1e-4)
+opt = Flux.setup(Adam(1f-4), (localization_net, classifier))
```
mcabbott, line 183: Not sure if this is too surprising. Rather than make a state tree per model, it makes one for the Tuple, and then calls `Flux.update!(opt, (localization_net, classifier), grads)` on both in one go.

ToucheSir: To reduce the surprise, we could destructure `grads` as well and re-tuple for the `update!` here. Slightly more verbose but more explicit.

mcabbott: In the VAE I just made two optimisers etc, like so:

https://github.com/FluxML/model-zoo/pull/375/files#diff-d2fcc7489d82937b5b466b70d629d200f23003b00a0e923cfa5031eedfa54d4cR112-R113

Maybe this wants a comment saying you could also do that?

vision/spatial_transformer/spatial_transformer.jl
```diff
@@ -174,16 +174,16 @@ classifier =
         Dense(256, 10),
     ) |> dev
 
-ps = Flux.params(localization_net, classifier)
+# ps = Flux.params(localization_net, classifier)
```
mcabbott, line 177:
```suggestion
```

vision/spatial_transformer/spatial_transformer.jl
```diff
@@ -79,28 +79,28 @@ end
...
     progress_tracker = Progress(length(train_loader), 1, "Training epoch $epoch :)")
     losses = zeros(length(train_loader))
     for (i, (x, y)) in enumerate(train_loader)
-        loss, grad = withgradient(ps) do
-            model_loss(x, y)
+        loss, grads = withgradient(localization_net, classifier) do ln, cl
+            model_loss(localization_net, classifier, x, y)
         end
-        Flux.update!(opt, ps, grad)
+        Flux.update!(opt, (localization_net, classifier), grads)
```
mcabbott, line 103:
```suggestion
        # Both the optimisers state `opt` and the gradients match a tuple of the two networks: 
        Flux.update!(opt, (localization_net, classifier), grads)
```

vision/spatial_transformer/spatial_transformer.jl
ToucheSir: The [PyTorch version](https://pytorch.org/tutorials/intermediate/spatial_transformer_tutorial.html) uses a single net, so if anything it seems like we should bundle them together more tightly. Maybe a custom layer?

mcabbott: I'm going to call that out of scope for this PR though. Just adding explicit gradients, without changing too much. Maybe it's good enough...

PR #385, opened 2022-12-30
Fix mlp mnist explicit gradients

### LINKED ISSUE #384: mlp_mnist can not be instantiated
When using the lastest Julia v1.6 (as listed in `compat` section of `Project.toml`) (currently 1.6.7), instantiating packages fails with cryptic message.

When deleting `Manifest.toml` first, instantiating works.

noxthot:
Fixes #384 and updates `mlp_mnist` to use explicit gradients (compat updated to use `Flux = 0.13.9` and `MLDatasets = 0.7`

### PR Checklist

- [ ] Tests are added -> not needed; manually executed with Julia LTS `1.6.7`
- [x] Documentation, if applicable

### DIFF
--- a/vision/mlp_mnist/mlp_mnist.jl
+++ b/vision/mlp_mnist/mlp_mnist.jl
@@ -89,7 +89,7 @@ end
 # Note that we use the functions [Dense](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Dense) so that our model is *densely* (or fully) connected and [Chain](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Chain) to chain the computation of the three layers.
 
 # ## Loss function
-loss(model, x, y) = logitcrossentropy(model(x), y)
+const loss = logitcrossentropy
 
 # Now, we define the loss function `loss_and_accuracy`. It expects the following arguments:
 # * ADataLoader object.
@@ -103,7 +103,7 @@ function loss_and_accuracy(data_loader, model, device)
     for (x, y) in data_loader
         x, y = device(x), device(y)
         ŷ = model(x)
-        ls += loss(model, x, y)
+        ls += loss(ŷ, y, agg=sum)
         acc += sum(onecold(ŷ) .== onecold(y)) ## Decode the output of the model
         num +=  size(x)[end]
     end
@@ -145,7 +145,7 @@ function train(; kws...)
     for epoch in 1:args.epochs
         for (x, y) in train_loader
             x, y = device(x), device(y) ## transfer data to device
-            gs = gradient(model -> loss(model, x, y), model) ## compute gradient
+            gs = gradient(m -> loss(m(x), y), model) ## compute gradient of the loss
             Flux.Optimise.update!(opt, model, gs[1]) ## update parameters
         end
 
--- a/vision/mlp_mnist/mlp_mnist.jl
+++ b/vision/mlp_mnist/mlp_mnist.jl
@@ -42,7 +42,9 @@ end
 function getdata(args)
     ENV["DATADEPS_ALWAYS_ACCEPT"] = "true"
 
-    ## Load dataset	
+    @info "Getting and transforming data"
+
+    ## Load dataset
     xtrain, ytrain = MLDatasets.MNIST(:train)[:]
     xtest, ytest = MLDatasets.MNIST(:test)[:]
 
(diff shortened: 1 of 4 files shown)


vision/mlp_mnist/mlp_mnist.jl
```diff
@@ -100,16 +103,16 @@ function loss_and_accuracy(data_loader, model, device)
     for (x, y) in data_loader
         x, y = device(x), device(y)
         ŷ = model(x)
-        ls += logitcrossentropy(ŷ, y, agg=sum)
+        ls += loss(model, x, y)
```
mcabbott, line 106: This runs the model twice, you could delete the `ŷ = model(x)` line.

vision/mlp_mnist/mlp_mnist.jl
```diff
@@ -131,21 +134,21 @@ function train(; kws...)
...
+    @info "Constructing model and starting training"
     ## Construct model
     model = build_model() |> device
-    ps = Flux.params(model) ## model's trainable parameters
-    
+
     ## Optimizer
-    opt = ADAM(args.η)
-    
+    opt = Flux.setup(ADAM(args.η), model)
```
mcabbott, line 142:
```suggestion
    opt = Flux.setup(Adam(args.η), model)
```

vision/mlp_mnist/mlp_mnist.jl
```diff
@@ -131,21 +134,21 @@ function train(; kws...)
...
+
     ## Training
     for epoch in 1:args.epochs
         for (x, y) in train_loader
             x, y = device(x), device(y) ## transfer data to device
-            gs = gradient(() -> logitcrossentropy(model(x), y), ps) ## compute gradient
-            Flux.Optimise.update!(opt, ps, gs) ## update parameters
+            gs = gradient(model -> loss(model, x, y), model) ## compute gradient
```
mcabbott, line 148: I think it's helpful to use different variable names for the local variable in the function:

```suggestion
            gs = gradient(m -> loss(m, x, y), model) ## compute gradient
```
Alternatively, having defined this function `loss`, you can just call it:
```suggestion
            gs = gradient(loss, model, x, y)  ## compute gradient
```

mcabbott (approved):
Looks good, modulo some minor points. Thanks!

vision/mlp_mnist/mlp_mnist.jl
noxthot: Thanks for your review, @mcabbott .

I have noticed this, but unfortunately `ŷ` is used in line 107 (new) to calculate the accuracy.

Thought it's better to have one centralized definition of loss than having `model(x)` executed twice.

CarloLucibello: just keep the previous definition
`ls += logitcrossentropy(ŷ, y, agg=sum)`
Also notice that the aggregation here is `sum`, not the default `mean`

mcabbott: Ah I see. 

Another style would be this: Just delete the definition `loss(model, x, y) = logitcrossentropy(model(x), y)` completely, and write the pieces both here and inside the gradient call. Maybe that's more explicit? 

If desired, it could have `const loss = logitcrossentropy` somewhere, so that there's one place to change it. But hard-coding it seems fine.

vision/mlp_mnist/mlp_mnist.jl
mcabbott: Or if you delete `loss` definition:
```suggestion
            gs = gradient(m -> logitcrossentropy(m(x), y), model)  ## compute gradient of the loss
```

vision/mlp_mnist/mlp_mnist.jl
noxthot: Thanks again for your comments.

@CarloLucibello: I restored aggregation by `sum` again (I changed this on purpose before to have the same loss used at every place; but I did not check carefully and actually the summed loss was divided by the number of samples afterwards. Thus it also calculated the `mean` within `loss_and_accuracy` and my previous commit introduced a bug.

@mcabbott: I decided for [this commit](https://github.com/FluxML/model-zoo/pull/385/commits/dacacb9859fbbe327edafe32d2a38d96a3a5cd6b) to account for your and @CarloLucibello's suggestions. Hope that's fine now and sorry for the temporarily introduced mini-bug regarding loss output.

vision/mlp_mnist/mlp_mnist.jl
noxthot: Thanks, @mcabbott.

Cf. https://github.com/FluxML/model-zoo/pull/385#discussion_r1059776647

vision/mlp_mnist/mlp_mnist.jl
mcabbott: Done, thanks for fixing this up!

PR #387, opened 2023-02-01
Update transfer learning tutorial

christiangnrd:
I updated the transfer learning tutorial for current Flux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Flux v0.13 is the latest right now, marked with ☀️; models upgraded to use  
 **Tutorials**
 * [A 60 Minute Blitz](tutorials/60-minute-blitz/60-minute-blitz.jl) ⛅️ v0.11
 * [DataLoader example with image data](tutorials/dataloader) ⛅️ v0.11
-* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ⛅️ v0.13
+* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ☀️ v0.13
 
 ## Examples Elsewhere
 
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Flux v0.13 is the latest right now, marked with ☀️; models upgraded to use  
 **Tutorials**
 * [A 60 Minute Blitz](tutorials/60-minute-blitz/60-minute-blitz.jl) ⛅️ v0.11
 * [DataLoader example with image data](tutorials/dataloader) ⛅️ v0.11
-* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ⛅️ v0.11
+* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ⛅️ v0.13
 
 ## Examples Elsewhere
 
--- a/tutorials/transfer_learning/Project.toml
+++ b/tutorials/transfer_learning/Project.toml
@@ -7,5 +7,5 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
-Flux = "0.11.5"
-julia = "1.5"
+Flux = "0.13"
+julia = "1.6"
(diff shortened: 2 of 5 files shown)


tutorials/transfer_learning/dataloader.jl
```diff
@@ -16,7 +16,7 @@ const CATS = filter(x -> occursin("cat", x), FILES)
 # Takes in the number of requested images per batch ("n") and image size
 # Returns a 4D array with images and an array of labels
 function load_batch(n = 10, nsize = (224,224); path = PATH)
-  if ((batchsize % 2) != 0)
+  if ((n % 2) != 0)
```
mcabbott, line 19: Perhaps clearer to use the named function?
```suggestion
  if isodd(n)
```

tutorials/transfer_learning/Project.toml
```diff
@@ -7,5 +7,5 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
-Flux = "0.11.5"
-julia = "1.5"
+Flux = "0.13"
```
mcabbott, line 10: Can you update the README to have `☀️ v0.13`, here:

https://github.com/christiangnrd/model-zoo/blob/transfer-fix/README.md?plain=1#L97

tutorials/transfer_learning/transfer_learning.jl
```diff
@@ -48,13 +48,15 @@ include("dataloader.jl")
 # Finally, the model looks something like:
 
 model = Chain(
-  resnet[1:end-2],
+  resnet[1],
+  AdaptiveMeanPool((1, 1)),
+  Flux.flatten,
   Dense(2048, 1000),  
   Dense(1000, 256),
```
mcabbott, line 55: It seems very odd to stack Dense layers with no nonlinearity. Would this still work?
```suggestion
  Dense(2048, 1000, relu),  
  Dense(1000, 256, relu),
```

tutorials/transfer_learning/transfer_learning.jl
```diff
@@ -65,7 +67,7 @@ opt = ADAM()
 loss(x,y) = Flux.Losses.logitcrossentropy(model(x), y)
 
 # Now to train
-# As discussed earlier, we don’t need to pass all the parameters to our training loop. Only the ones we need to
+# As discussed earlier, we don't need to pass all the parameters to our training loop. Only the ones we need to
 # fine-tune. Note that we could have picked and chosen the layers we want to train individually as well, but this
 # is sufficient for our use as of now.
```
mcabbott, line 73: I can't comment below, probably this should remove `@epochs` in favour of `for epoch in 1:2`.

A bigger change would be to move on from implicit parameters, `params` etc, but perhaps that can be the next PR.

mcabbott (commented):
Thanks!

tutorials/transfer_learning/Project.toml
christiangnrd: Done

tutorials/transfer_learning/transfer_learning.jl
christiangnrd: I also found it weird to stack linear layers but my goal was to get it working so that I had an example to base it off of and pushing the fix was a bonus.

It seems to work with the relu.

tutorials/transfer_learning/transfer_learning.jl
christiangnrd: I made the basic fix. I don't think it's worth making the explicit parameter changes until Metalhead gets an update as the current Metalhead stable version's compat prevents Flux from updating past 0.13.4 because of the maximum versions for Functors and MLUtils.

README.md
```diff
@@ -94,7 +94,7 @@ Flux v0.13 is the latest right now, marked with ☀️; models upgraded to use
 **Tutorials**
 * [A 60 Minute Blitz](tutorials/60-minute-blitz/60-minute-blitz.jl) ⛅️ v0.11
 * [DataLoader example with image data](tutorials/dataloader) ⛅️ v0.11
-* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ⛅️ v0.11
+* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ⛅️ v0.13
```
mcabbott, line 97:
```suggestion
* [Transfer Learning](tutorials/transfer_learning/transfer_learning.jl) ☀️ v0.13
```

mcabbott (approved):

PR #388, opened 2023-02-06
Fixed the tutorials link to the correct page in ReadMe

Ochibobo:
[Please delete this text and describe your change here.
For bugfixes, please detail the bug and include a test case which your patch fixes.
If you are adding a new feature, please clearly describe the design, its rationale, the possible alternatives considered.
It is easiest to merge new features when there is clear precedent in other systems; we need to know we're taking
the right direction since it can be hard to change later.]

### PR Checklist

- [ ] Tests are added
- [ ] Documentation, if applicable


ToucheSir (approved):
Thanks!

PR #394, opened 2023-02-11
Simplify LeNet

mcabbott:
There's a lot of junk in these files. Perhaps we should clean them up.

Text can go in README. Code should encourage interactive use, and hence avoid big functions called exactly once. No more parameter structs. 

Readable version, since the diff is huge:  https://github.com/FluxML/model-zoo/blob/simpler_lenet/vision/conv_mnist/conv_mnist.jl

~~WIP, would like https://github.com/FluxML/Flux.jl/pull/2186 and https://github.com/JuliaML/MLUtils.jl/issues/144 .~~ Edit: or just https://github.com/FluxML/Flux.jl/pull/2240 now.

### DIFF
--- a/vision/conv_mnist/conv_mnist.jl
+++ b/vision/conv_mnist/conv_mnist.jl
@@ -56,14 +56,14 @@ lenet = Chain(
 
 y1hat = lenet(x1)  # try it out
 
-softmax(y1hat)
+sum(softmax(y1hat); dims=1)
 
 # Each column of softmax(y1hat) may be thought of as the network's probabilities
 # that an input image is in each of 10 classes. To find its most likely answer, 
 # we can look for the largest output in each column, without needing softmax first. 
 # At the moment, these don't resemble the true values at all:
 
-hcat(Flux.onecold(y1hat, 0:9), Flux.onecold(y1, 0:9))
+@show hcat(Flux.onecold(y1hat, 0:9), Flux.onecold(y1, 0:9))
 
 #===== METRICS =====#
 
@@ -80,7 +80,7 @@ function loss_and_accuracy(model, data::MNIST=test_data)
     (; loss, acc, split=data.split)  # return a NamedTuple
 end
 
-loss_and_accuracy(lenet)  # accuracy about 10%
+@show loss_and_accuracy(lenet);  # accuracy about 10%, before training
 
 #===== TRAINING =====#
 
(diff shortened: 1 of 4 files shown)


vision/conv_mnist/conv_mnist.jl
```diff
@@ -1,240 +1,215 @@
...
+    if epoch % 2 == 1
+        loss, acc, _ = loss_and_accuracy(lenet)
+        test_loss, test_acc, _ = loss_and_accuracy(lenet, test_data)
+        @info "logging:" epoch acc test_acc
+        nt = (; epoch, loss, acc, test_loss, test_acc)  # make a NamedTuple
+        push!(train_log, nt)
+    end
+    if epoch % 5 == 0
+        filename = joinpath("runs", "lenet.jld2")
+        JLD2.jldsave(filename; lenet_state = Flux.state(lenet |> cpu))
```
mcabbott, line 116: I changed this to save using `state` and JLD2.

Unlike BSON, what this won't do is make the folder to write into. Not sure what's best. Could make it manually. Or just change gitignore to have `*.jld2`?

mcabbott: [b32a41c](https://github.com/FluxML/model-zoo/pull/394/commits/b32a41c35b943bf37cfecad26375582d5c7c9074) makes the folder by hand

vision/conv_mnist/conv_mnist.jl
```diff
@@ -1,240 +1,218 @@
...
+loaded_state = JLD2.load(filename, "lenet_state");
+
+# Now you would normally re-create the model, and copy all parameters into that.
+# We can use lenet2 from just above:
+
+Flux.loadmodel!(lenet2, loaded_state)
+
+# Check that it now agrees with the earlier, trained, model:
+
+lenet2(cpu(x1)) ≈ cpu(lenet(x1))
```
ToucheSir, line 215: Just repeating the suggestion in https://github.com/FluxML/model-zoo/pull/397#discussion_r1185650292 to print some of these out so that users running the script directly aren't completely left without feedback.

mcabbott: See what you think of [9bf6c63](https://github.com/FluxML/model-zoo/pull/394/commits/9bf6c636917dc9913abaf6c56c31c4244dfd8dd3), adds more `@show`.

It doesn't print many headings etc. when run as a script. Could add more but it's all extra clutter. If you run this as a script, I think you need the source open in another window to learn anything anyway.

PR #397, opened 2023-04-25
Simpler MLP

mcabbott:
Same idea as https://github.com/FluxML/model-zoo/pull/394, to remove all this baroque junk of making structs you use once, and functions you call once, to just focus on the essentials. But even simpler than #394, no GPU, no NamedTuples, no saving. 

* Before: https://github.com/FluxML/model-zoo/blob/master/vision/mlp_mnist/mlp_mnist.jl

* After: https://github.com/FluxML/model-zoo/blob/simpler_mlp/vision/mlp_mnist/mlp_mnist.jl

~~Maybe some of the text in the old one should move to the readme.~~

### DIFF
--- a/vision/mlp_mnist/README.md
+++ b/vision/mlp_mnist/README.md
@@ -6,7 +6,7 @@
 
 ## Model Info
 
-A [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) (MLP) consists of at least three sets of of nodes: an input layer, a hidden layer and an output layer. Except for the input node each node is a neuron that uses a nonlinear activation function. MLP utilizes a supervised learning technique called backpropagation for training. Its multiple layers and non-linear activation distinguish MLP from a linear perceptron. It can distinguish data that is not linearly separable.
+A [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) (MLP) consists of at least three sets of of nodes: an input layer, a hidden layer and an output layer. Each node except for the input node is a neuron that uses a nonlinear activation function. The multiple layers and non-linearities allow an MLP to distinguish data that is not linearly separable once trained.
 
 In this example, we create a MLP that classifies handwritten digits using the [MNIST dataset](http://yann.lecun.com/exdb/mnist/).
 Our model uses the simplest [Flux layers](http://fluxml.ai/Flux.jl/stable/models/layers/), namely `Dense` and `Chain`.
--- a/vision/mlp_mnist/README.md
+++ b/vision/mlp_mnist/README.md
@@ -6,7 +6,7 @@
 
 ## Model Info
 
-A [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) (MLP) consists of at least three sets of of nodes: an input layer, a hidden layer and an output layer. Each node except for the input node is a neuron that uses a nonlinear activation function. The multiple layers and non-linearities allow an MLP to distinguish data that is not linearly separable once trained.
+A [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) (MLP) consists of at least three sets of of nodes: an input layer, one or more hidden layer and an output layer. Each node except for the input node is a neuron that uses a nonlinear activation function. The multiple layers and non-linearities allow an MLP to distinguish data that is not linearly separable once trained.
 
 In this example, we create a MLP that classifies handwritten digits using the [MNIST dataset](http://yann.lecun.com/exdb/mnist/).
 Our model uses the simplest [Flux layers](http://fluxml.ai/Flux.jl/stable/models/layers/), namely `Dense` and `Chain`.
--- a/vision/mlp_mnist/README.md
+++ b/vision/mlp_mnist/README.md
@@ -6,9 +6,19 @@
 
 ## Model Info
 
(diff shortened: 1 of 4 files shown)


vision/mlp_mnist/mlp_mnist.jl
```diff
@@ -1,174 +1,116 @@
...
-# ## Train function
+using ImageCore, ImageInTerminal
 
-# Now, we define the `train` function that calls the functions defined above and trains the model.
+xtest, ytest = only(simple_loader(test_data, batchsize=length(test_data)));
 
-function train(; kws...)
-    args = Args(; kws...) ## Collect options in a struct for convenience
+# There are many ways to look at images, you won't need ImageInTerminal if working in a notebook
+# ImageCore.Gray is a special type, whick interprets numbers between 0.0 and 1.0 as shades:
```
ToucheSir, line 106:
```suggestion
# There are many ways to look at images, you won't need ImageInTerminal if working in a notebook.
# ImageCore.Gray is a special type, which interprets numbers between 0.0 and 1.0 as shades:
```

vision/mlp_mnist/mlp_mnist.jl
```diff
@@ -1,174 +1,116 @@
...
-    acc = 0
-    ls = 0.0f0
-    num = 0
-    for (x, y) in data_loader
-        x, y = device(x), device(y)
-        ŷ = model(x)
-        ls += loss(ŷ, y, agg=sum)
-        acc += sum(onecold(ŷ) .== onecold(y)) ## Decode the output of the model
-        num +=  size(x)[end]
+    if mod(epoch, 2) == 1
```
ToucheSir, line 88:
```suggestion
    if isodd(epoch)
```
For your consideration.

vision/mlp_mnist/README.md
```diff
@@ -6,10 +6,20 @@
 
 ## Model Info
 
-An MLP consists of at least three of nodes: an input layer, a hidden layer and an output layer. Except for the input node each node is a neuron that uses a nonlinear activation function. MLP utilizes a supervised learning technique called backpropagation for training. Its multiple layers and non-linear activation distinguish MLP from a linear perceptron. It can distinguish data that is not linearly separable.
+A [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) (MLP) consists of at least three sets of of nodes: an input layer, a hidden layer and an output layer. Except for the input node each node is a neuron that uses a nonlinear activation function. MLP utilizes a supervised learning technique called backpropagation for training. Its multiple layers and non-linear activation distinguish MLP from a linear perceptron. It can distinguish data that is not linearly separable.
```
ToucheSir, line 9:
```suggestion
A [multi-layer perceptron](https://en.wikipedia.org/wiki/Multilayer_perceptron) (MLP) consists of at least three sets of of nodes: an input layer, a hidden layer and an output layer. Each node except for the input node is a neuron that uses a nonlinear activation function. The multiple layers and non-linearities allow an MLP to distinguish data that is not linearly separable once trained.
```
Mostly fixing up the old version's use of articles and dropping the out-of-place sentence about backprop.

vision/mlp_mnist/README.md
```diff
@@ -6,10 +6,20 @@
...
+Our model uses the simplest [Flux layers](http://fluxml.ai/Flux.jl/stable/models/layers/), namely `Dense` and `Chain`.
+Since it uses [softmax](https://en.wikipedia.org/wiki/Softmax_function) on its outputs, and [`crossentopy`](http://fluxml.ai/Flux.jl/stable/models/losses/#Flux.Losses.crossentropy) as the loss function.
+
+For simplicity this model does not use a graphics card, since an ordinary CPU is fast enough.
+See for example the [LeNet convolutional network](https://github.com/FluxML/model-zoo/tree/master/vision/conv_mnist) for GPU usage.
 
 ## Training
 
+You can copy and paste the example into the Julia REPL to see what each part does.
+Or you can run it all at once from the terminal, like this:
```
ToucheSir, line 21: Running all at once from the terminal won't show any output. Could we use `print` on the predictions and display the image even if users aren't running from the REPL?

ToucheSir (commented):
One benefit of all the extra comments in the original was that it could be rendered as a Literate notebook. I wonder if we should make that official with either a README note or top-level comment with brief instructions.

vision/mlp_mnist/README.md
mcabbott: For me it does print:
```
mlp_mnist % /Applications/Julia-1.9.app/Contents/Resources/julia/bin/julia --project mlp_mnist.jl
┌ Info: After epoch = 1
│   loss = 1.8101638024672866
│   train_acc = 75.81
└   test_acc = 76.75
┌ Info: After epoch = 3
│   loss = 0.8674804433248937
│   train_acc = 85.65
└   test_acc = 86.22
```

vision/mlp_mnist/README.md
mcabbott: Sure. This is awful but I didn't want to kill it.

vision/mlp_mnist/mlp_mnist.jl
mcabbott: Originally I had a `print_every` variable, and then simplified to hard-code everything. I do think this way points out better how to do something every 10, etc.

mcabbott:
> original was that it could be rendered as a Literate notebook

I think quite a few of these files were rendered this way for the old website. But they got stale... and IDK how valuable rendering & commentary is anyway. 

This PR and its friend want to go the other direction. Having relatively few, concise, clear examples means you're more likely to find the right one, and it's more likely to be up to date.

vision/mlp_mnist/README.md
ToucheSir: If not the images, perhaps we could print out the predictions/ground truth labels in lines 110+115 for comparison.

ToucheSir (approved):

Issue #398, opened 2023-07-12
char-rnn example throws MethodError: no method matching ndims(::SubString{String})


markusgumbel:
When I run the **char-rnn** example from the model-zoo I get the following error (full stack trace at bottom):
```
ERROR: LoadError: MethodError: no method matching ndims(::SubString{String})
```
It happens in the `getdata` function while executing the following statements.
```
    Xs = partition(batchseq(chunk(text, args.batchsz), stop), args.seqlen)
    Ys = partition(batchseq(chunk(text[2:end], args.batchsz), stop), args.seqlen)
```
The problem seems to be that `chunk(text, args.batchsz)` creates an array of `SubString`s. `batchseq` calls `ndims` on this array. However, `ndims` does not seem to be defined for `SubString`s.
When I replace the lines with
```
    Xtext = [collect(t) for t in chunk(text, args.batchsz)]
    Ytext = [collect(t) for t in chunk(text[2:end], args.batchsz)]
    Xs = partition(batchseq(Xtext, stop), args.seqlen)
    Ys = partition(batchseq(Ytext, stop), args.seqlen)
```
it works. The Strings are converted into an array of characters (for which `ndims` is available).

My environment is:

- Ubuntu 22.04 LTS
- Julia 1.9.2

The program was started according to the documentation:
`julia --project char-rnn.jl` within the directory `text/char-rnn`.

Thanks a lot in advance for any help for this issue. And many thanks for the Flux framework in general!

Full stack trace:
```
ERROR: LoadError: MethodError: no method matching ndims(::SubString{String})

Closest candidates are:
  ndims(::CUDA.CUSPARSE.CuSparseMatrix)
   @ CUDA ~/.julia/packages/CUDA/Ey3w2/lib/cusparse/array.jl:225
  ndims(::CUDA.CUTENSOR.CuTensor)
   @ CUDA ~/.julia/packages/CUDA/Ey3w2/lib/cutensor/tensor.jl:23
  ndims(::CUDA.CUSPARSE.CuSparseDeviceMatrixCSC)
   @ CUDA ~/.julia/packages/CUDA/Ey3w2/lib/cusparse/device.jl:36
  ...

Stacktrace:
  [1] (::MLUtils.var"#142#146")(x::SubString{String})
    @ MLUtils ~/.julia/packages/MLUtils/KcBtS/src/utils.jl:427
  [2] mapreduce_first(f::MLUtils.var"#142#146", op::Function, x::SubString{String})
[...]

PR #406, opened 2024-02-04
PR for issue #398.

markusgumbel:
Fixed issue #398. 
 * Changed file `char-rnn.jl` according to description/sugesstion in issue text.
 * Set Julia version to 1.10. The Julia 1.10 runtime was not able to run the 1.6 version.

### DIFF
--- a/text/char-rnn/char-rnn.jl
+++ b/text/char-rnn/char-rnn.jl
@@ -66,8 +66,10 @@ function getdata(args::Args)
     
     ## Partitioning the data as sequence of batches, which are then collected 
     ## as array of batches
-    Xs = partition(batchseq(chunk(text, args.batchsz), stop), args.seqlen)
-    Ys = partition(batchseq(chunk(text[2:end], args.batchsz), stop), args.seqlen)
+    Xtext = [collect(t) for t in chunk(text, args.batchsz)]
+    Ytext = [collect(t) for t in chunk(text[2:end], args.batchsz)]
+    Xs = partition(batchseq(Xtext, stop), args.seqlen)
+    Ys = partition(batchseq(Ytext, stop), args.seqlen)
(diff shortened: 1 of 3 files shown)


text/char-rnn/char-rnn.jl
```diff
@@ -66,8 +66,10 @@ function getdata(args::Args)
     
     ## Partitioning the data as sequence of batches, which are then collected 
     ## as array of batches
-    Xs = partition(batchseq(chunk(text, args.batchsz), stop), args.seqlen)
-    Ys = partition(batchseq(chunk(text[2:end], args.batchsz), stop), args.seqlen)
+    Xtext = [collect(t) for t in chunk(text, args.batchsz)]
+    Ytext = [collect(t) for t in chunk(text[2:end], args.batchsz)]
+    Xs = partition(batchseq(Xtext, stop), args.seqlen)
+    Ys = partition(batchseq(Ytext, stop), args.seqlen)
```
ToucheSir, line 72: Can you explain the reasoning behind this change? Generally, I would not expect `collect` to be required here.

markusgumbel: The problem seems to be that `chunk(text, args.batchsz)` creates an array of `SubString`s. `batchseq` calls `ndims` on this array. However, `ndims` does not seem to be defined for `SubString`s. The Strings are converted into an array of characters (for which `ndims` is available).
Without the change you get:
ERROR: LoadError: MethodError: no method matching ndims(::SubString{String})

ToucheSir: I see. We should fix `batchseq` to work with strings, but that can happen separately. Thanks for elaborating.

ToucheSir (approved):
Thanks for updating this!

markusgumbel:
Thank you, too.

PR #407, opened 2024-03-18
revamp vae_mnist

CarloLucibello:
A light revamp, to keep up with recent changes (e.g. use JLD2 instead of BSON)


CarloLucibello:
merge?

mcabbott:
Oh no I see I was fiddling offline with the same model, and when I finally got it done I saw this PR. 

We did some of the same things. Differences:
* You use `randn_like(logσ)`, I did `.* randn.(Float32)`
* You kept the `mutable struct Args` and function train, I killed them. Thoughts? I don't know why all these models do that, but maybe I missed something. 
* You use `Flux.get_device()`, I left `|> gpu`, which should be the same?
* I didn't touch `vae_plot.jl`, has to be updated not to use `BSON.@load`

CarloLucibello:
> you use `randn_like(logσ)`, I did `.* randn.(Float32)`

is it the same on device? I prefer `randn_like` though since it is more transparent

> You kept the mutable struct Args and function train, I killed them. Thoughts? I don't know why all these models do that, but maybe I missed something.

Args was introduced in order to have both script usage `julia vae_minst.jl` and interactive usage `julia> train(eta=1e-3)`. 
The `train` function is for enforcing the generically good julia practice of not having globals. I guess we can get rid of both.

mcabbott:
Let me merge this and work on top... see if I can combine things.

PR #408, opened 2024-03-20
updated VAE

mcabbott:
Originally this just wanted to add `@layer` since this is the simplest model with a `struct`...


CarloLucibello:
Nooo
Take what you want from the other PR, I'll close it

